Issue #25, opened 2016-01-05
Morganey DSL inside of Scala


rexim:
For quick AST definition inside of Scala.

rexim:
Resolved by 435dd45c9940e8805e9d22927dcb15f83f436deb

PR #61, opened 2016-07-01
Implemented quasiquotations

keddelzz:
## The birth of morganey.meta._

Implemented quasiquotations. Now it's possible to create and interpolate lambda terms with a string-interpolator. Lifting of Scala-values to lambda terms is also possible.

``` scala
val two  = 2
val three = m"3"

val eight = m"( $two $three )"
```

Unquoting of lambda terms allows detection of complex patterns. And unlifting even allows the automatic conversion of lambda terms back to Scala:

``` scala
val term: LambdaTerm = ...

term match {
  case m"( ${a: Int} ${b: Int} )" => m"${math.pow(a, b).toInt}"
  case _ => ???
}
```

This feature was implemented in scope of #25


rexim:
Looks cool! I'm gonna review this during the day and try to merge the PR before Sunday. Thanks! :)

keddelzz:
Should I describe the implemented features in more Detail?

rexim:
@keddelzz, The detailization seems to be alright for me. If something is not clear I will ask question in the comments. :) Right now I'm a bit busy. I will take a look into the PR once I'm home. In 5-6 hours or so.

@ForNeVeR, I saw you showed the interest in this PR too. Your comments are appreciated. :)

build.sbt
```diff
@@ -1,17 +1,61 @@
...
+lazy val macroSettings = Seq(
+  name := "morganey-macros",
+  libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value
+)
 
-scalacOptions ++= Seq("-unchecked", "-deprecation", "-feature")
+lazy val kernelSettings = Seq(
+  name := "morganey-kernel",
+  libraryDependencies += "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4"
+)
```
rexim, line 30: I'm not an expert in macros, but why was it so essential to split the project into subprojects to be able to work with macros? Don't get me wrong I like the idea. :) `meta` works above `kernel` and `kernel` knows nothing about `meta`. This looks right to me. But still what was the motivation to split the project? Was this something related to the technical limitations of macros or just a project structure improvement?

kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
```diff
@@ -14,7 +14,9 @@ object IntMatcher {
     }
 }
 
-object LambdaParser extends RegexParsers {
+object LambdaParser extends LambdaParser
```
rexim, line 17: Is this companion object really required here?

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
```diff
@@ -0,0 +1,161 @@
+package me.rexim.morganey.meta
+
+import me.rexim.morganey.ast._
+
+import scala.collection.mutable.ArrayBuffer
+import scala.reflect.macros.whitebox.Context
+
+private[meta] class QuotationMacro(val c: Context) {
```
rexim, line 8: That's a pretty cool class. :) I believe this is where all of the macro magic happens. But it has a lot of imperative logic, which I'd like to cover with any kind of tests that are run on CI. Is it possible to do that quickly? If not we can make it a separate issue and think about it later.

macros/src/main/scala/me/rexim/morganey/meta/package.scala
```diff
@@ -0,0 +1,19 @@
...
+  with DefaultUnliftableInstances {
+
+  private[meta] trait Quasiquotation {
+    def apply[T](args: T*): Any = macro QuotationMacro.quote
+    def unapply(arg: Any): Any = macro QuotationMacro.unquote
+  }
+
+  implicit class Quasiquote(ctx: StringContext) {
+    object m extends Quasiquotation
+    object lc extends Quasiquotation
```
rexim, line 16: What do we need the `lc` interpolator for? It looks like it does the same thing as `m`, no?

rexim:
Currently I actually like what I see. And the quasiquotation seems to be working as expected on my side locally! But I still don't quite understand the `QuotationMacro` class. Mostly because it's almost 3:00am for me! :D

@keddelzz, I'm gonna continue my review tomorrow if you don't mind. I'll try to finish it before the Sunday stream.

Thanks!

build.sbt
keddelzz, line 30: Macro Implementations have to be in a separate projects. They'll have to be compiled before the code, which uses the macro.
Because the macro uses your Parser and the definitions of AST-Nodes, the project structure has to be split. It might be possible to join the definitions of lambda-terms, parser and macros into one subproject, but macros and main-program (for example repl-implementation) have to be separate.

kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
keddelzz, line 17: Inheritance from your parser was necessary for code reuse (since the interpolator also requires a parser, but with the capability for parsing special names for holes used in interpolation).

The companion object is used by your code in the Main-class. To not break your repl I kept the api which was exposed by your parser.

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
keddelzz, line 8: It would be possible to test the result of a macro expansion, but I think it's impossible to run tests while expansion. I'm not quite sure, though. We'd have to try it.

If you want, you could create an issue and assign it to me, I'll add some tests for expansion-results later.

macros/src/main/scala/me/rexim/morganey/meta/package.scala
keddelzz, line 16: Yes, the lc interpolator does completely do the same thing as m. So if you want we could remove it. I thought it would nice to choose between which interpolator to use.

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
ForNeVeR, line 8: I think that the macro library is an ordinary sbt project and it actually can be tested without involving our main project; we could call that our _unit_ tests. At the same time, we can and should also test the results of the macro-expansion, that will be form our _integration_ test layer.

Personally I think that it's preferable to include tests (at least integration ones) in the same PR as the main code. But it'd be okay to write them a bit later. Let @rexim to choose.

macros/src/main/scala/me/rexim/morganey/meta/package.scala
ForNeVeR, line 16: Aha, I got that. `m` stands for _Morganey_ and `lc` stands for _Lambda Calculus_. I think it's better to leave only `lc`. Single-letter interpolators is standard library's prerogative I'd say.

keddelzz, line 16: I think `m` would be a better name, because it's a reference of the projects name. It would be possible to rename it, to some longer name: `mor` or `mg`, or something like that. If you really wanted it could also be renamed to `morganey`, but I think that'll be too long for the name for an interpolator.

I'm not quite sure why you think single-letter interpolators is standard library's prerogative, after all Scala let's you give them arbitrary names .

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
keddelzz, line 8: The macro project can of course be unit-tested separately, but I think It's impossible to test the macro-code itself (class QuotationMacro). As of now [SIP-5989](https://issues.scala-lang.org/browse/SI-5989) does still exist, which addresses the (unit-)testability of macros.

ForNeVeR, line 8: Well, okay then. Our main focus would be on testing the expanded macros anyways.

macros/src/main/scala/me/rexim/morganey/meta/package.scala
keddelzz, line 16: Other names i've thought of: `lambda`, `ast` and `tree`. I didn't like all of those ideas, because names were either too long for my taste (for interpolators) or they were not related to the project. 

Ultimately @rexim has to decide, which name he wants.

keddelzz:
Unit tests might have to be rearranged. For now tests for parser and lambda terms are defined in subproject `core`, but they do belong to `kernel`.

build.sbt
rexim, line 30: Got it. Thanks for the explanation!

kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
rexim, line 17: Ah, yeah! Right. :) Makes sense. Thanks!

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
rexim, line 8: Yeah, I'm gonna create a separate issue and we will discuss this later when the time is right.

macros/src/main/scala/me/rexim/morganey/meta/package.scala
rexim, line 16: I'm not against using `m` or `lc` or both simultaneously. I was just curious why do we have two interpolators that do the same thing. Now I see what they mean. I'm ok with providing them both to the user as available options.

rexim:
@keddelzz,

> Unit tests might have to be rearranged. For now tests for parser and lambda terms are defined in subproject `core`, but they do belong to `kernel`.

Alright, please do that if it can be done quickly. If not as usual let's just create an issue to not forget about that and move forward. :)

keddelzz:
I'll do it as fast as I can!

rexim:
@keddelzz I mean, no pressure on that or anything. It's completely optional! I really appreaciate your work! I just want to get your PR merged before 10pm NOVT (in 4 hours) to actually start using the feature on the stream. :) I respect your personal time and if it can't be done before the stream it's totally fine! Really thank you for all of your work!

rexim:
Alright, I more or less understand what is happening in the `QuotationMacro` class and I'm ready to merge this PR. Gonna do that right on the stream. :)

keddelzz:
I've split the tests to their appropriate projects. All builds are passing.

If you want I could write more (integration) tests for macro expansion results later, but this will have to wait a bit. Maybe I could add some in the next week.

rexim:
@keddelzz alright! Looks good! Thank you very much! I'm merging this now then.

PR #69, opened 2016-07-04
Tests for macros in meta subproject

### LINKED ISSUE #63: Testing macros in meta subproject on CI
Derived from discussion: https://github.com/rexim/Morganey/pull/61#discussion_r69382326

keddelzz:
Close #63

### DIFF
--- a/kernel/src/test/scala/me/rexim/morganey/ChurchNumberConverterSpec.scala
+++ b/kernel/src/test/scala/me/rexim/morganey/ChurchNumberConverterSpec.scala
@@ -1,15 +1,10 @@
 package me.rexim.morganey
 
 import me.rexim.morganey.church.ChurchNumberConverter._
-import me.rexim.morganey.ast.LambdaTermHelpers._
 import me.rexim.morganey.helpers.TestTerms
 import org.scalatest._
 
 class ChurchNumberConverterSpec extends FlatSpec with Matchers with TestTerms {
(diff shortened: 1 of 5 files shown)


rexim:
LGTM. Thanks!

keddelzz:
I did some research why 

``` scala
m"\"Hello\""
```

is not working, I found out, that I treated escape sequences incorrectly.
(This problem did arise on your Sunday-stream.)

So I fixed this in the scope of this PR (59d3f0e). I'm sorry to say that the example shown above still doesn't work, because of a bug in scala ([SI-6476](https://issues.scala-lang.org/browse/SI-6476)). But \n, \t, \' etc are working correct now.

rexim:
@keddelzz thank you for putting so much time and effort into that! It's totally fine if such functionality doesn't work right now because of a bug in the compiler. It's not critical. I'd suggest to create an issue for that. Something like "Escaped double quotes don't work in quasiquotation" and refer to SI-6476. Once the compiler's bug is fixed we gonna resolve our issue too.

The presence of such issue will show that we are aware of this bug, will serve as a reminder for ourselves and also as a place where people can complain if they encounter this problem during using Morganey. :)

keddelzz:
Morganey is such a fun project, I'm totally hyped and very much interested in it. So of course I'll invest time in something I'm interested in. :)

PR #76, opened 2016-07-09
Fix nullpointer exception on ^D

### LINKED ISSUE #75: Fix nullpointer exception on ctrl-d in repl

keddelzz:
Close #75

### DIFF
--- a/src/main/scala/me/rexim/morganey/Main.scala
+++ b/src/main/scala/me/rexim/morganey/Main.scala
@@ -63,19 +63,15 @@ object Main extends SignalHandler {
     val con = new ConsoleReader()
     con.setPrompt("λ> ")
 
-    def line(): String = {
-      val rawLine = con.readLine()
-      if (rawLine eq null) null
-      else rawLine.trim()
-    }
+    def line() = Option(con.readLine()).map(_.trim)
 
     val evalLine = handleLine(con) _
 
     while (running) line() match {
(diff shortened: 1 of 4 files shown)


src/main/scala/me/rexim/morganey/Main.scala
```diff
@@ -29,37 +29,54 @@ object Main extends SignalHandler {
...
     val con = new ConsoleReader()
     con.setPrompt("λ> ")
 
-    while (true) {
-      val line = con.readLine().trim()
+    def line(): String = {
+      val rawLine = con.readLine()
+      if (rawLine eq null) null
+      else rawLine.trim()
+    }
```
rexim: This is fine, but how about this?

```
def line(): String =
  Option(con.readLine())
    .map(_.trim)
    .getOrElse(null)
```

I think it's more scala-ish :) Or, if the point to only check for null result in the while-loop below we can make it even simplier:

```
def line(): Option[String] = Option(con.readLine()).map(_.trim)
```

And check for `None` instead of `null` in the loop.

What do you think?

keddelzz: I think with Option it's better! I'll change it later!

rexim: Cool! Thank you!

rexim:
LGTM! Thank you very much!

PR #77, opened 2016-07-10
Module system

### LINKED ISSUE #73: Extensible standard library / module system
It would be nice, if Morganey knew some basic operations like `succ`, `pred`, `id`, `first`, `second`... Therefore I suggest the implementation of a standard library, which can be extended by the user of morganey.

My idea is, that there's a folder somewhere on the computer where you can add morganey-source-files and they are automatically loaded into each REPL-session and as a library for each program.

Instead of being loaded automatically the modules could also be loaded via an import/include statement.

rexim:
Things left to implement:
- Standard library #78 
- Prelude #79 
- Cycled loads #80 

Everything has a separate issue. Hence close #73


rexim:
This is pretty much everything what I've managed to get done on the stream. It seems to be working but I want to polish it a bit. Plan to finish that tomorrow.

Also, there is a couple of things unimplemented (like standard library, prelude, etc.). I'm gonna make separate issues for those later.

rexim:
I think I'm done on that.

@keddelzz, @ForNeVeR, could you guys please review my changes. Your comments are appreciated.

README.md
```diff
@@ -18,14 +18,16 @@ To run the REPL just enter the following in the source code directory
...
+    <repl-command> := <term> | <binding> | <loading>
     <term> ::= <variable>
              | <numeric-literal>
              | <character-literal>
              | <string-literal>
              | <function>
              | <application>
     <binding> ::= <variable> := <term>
+    <loading> ::= load <module-path>
+    <module-path> := [a-zA-Z][a-zA-Z0-9/]*
```
ForNeVeR: I am sure it would be better to define `module-path` as `string-literal`. Your current definition doesn't supports paths with spaces or cyrillic (greek?) letters, that's not nice!

morganey.txt
```diff
@@ -1,6 +1,5 @@
-SUCC := (λn.(λf.(λx.(f ((n f) x)))))
-PLUS := (λm.(λn.(λf.(λx.((m f) ((n f) x))))))
-MULT := (λm.(λn.(λf.(m (n f)))))
+load arithmetic
+
 PAIR := (λx . (λy . (λ z . ((z x) y))))
 I := (λx . x)
 a0 := (λf . I)
```
ForNeVeR, line 5: Why don't rename this file to `morganey.morganey` or `hello.morganey` or `test.morganey` or something? It seems that you've chosen the `.morganey` as a standard file extension, so let's be consistent!

README.md
rexim: I'm not sure it would be better. I mean it's hard for me to call spaces or non-english letters in for instance java package names a good practice. I don't mind to support that though. But only as a separate issue, and only with the strong justification why it is really needed for someone. So go ahead and file one. We can continue discussing this there.

morganey.txt
rexim, line 5: Yep, seems to be a good idea. During implementing this I was not sure about the official Morganey extension. But looks like `.morganey` it is. Gonna rename that.

README.md
ForNeVeR: I believe it was meant as a filesystem path because no sane module system includes `/` as a namespace separator.

_(PHP **does** and that's the exact reason it's not sane)_

Maybe replace `/` with `.` there? It will state your intention more clear.

rexim: That is a better idea! Thanks! Let me make that happen...

morganey.txt
rexim, line 5: Done

README.md
rexim: Done

ForNeVeR:
LGTM :+1:

keddelzz:
LGTM for me, too!

rexim:
Cool! Thank you guys!

PR #86, opened 2016-07-13
Autocompletion in REPL

### LINKED ISSUE #70: Autocomplete in REPL
It would be cool to be able to tab autocomplete vars form the global context.

keddelzz:
Autocompletion for:
- global bindings
- paths in load-statements

Close #70

### DIFF
--- a/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
+++ b/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
@@ -65,9 +65,10 @@ class ReplAutocompletion(globalContext: () => List[MorganeyBinding]) extends Com
       if (isMorganeyModule(f)) new File(f.getParent, f.getName.replaceAll(s".$fileExtension", ""))
       else f
 
-    def everythingIn(path: List[String]) =
+    def everythingIn(path: List[String], fileNameFilter: String => Boolean = _ => true) =
       findAllModulesIn(path.mkString("."))
         .map { case (root, f) => (root, stripExtensionIfModuleFile(f)) }
+        .filter { case (root, f) => fileNameFilter(f.getName) }
         .map(relativize)
         .map(relativeFileToLoadPath)
 
@@ -84,7 +85,7 @@ class ReplAutocompletion(globalContext: () => List[MorganeyBinding]) extends Com
       // load |
       case (Nil, false)      => topLevelDefinitions().map(stripExtensionIfModuleFile).map(_.getName)
       // load math.ari|
-      case (xs :+ x, false)  => everythingIn(xs).filter(_ startsWith x)
+      case (xs :+ x, false)  => everythingIn(xs, _ startsWith x)
     }
   }
 
--- a/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
+++ b/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
@@ -1,11 +1,14 @@
 package me.rexim.morganey
 
+import java.io.File
 import java.util.{List => Jlist}
 
 import jline.console.completer.Completer
 import me.rexim.morganey.ast.MorganeyBinding
+import me.rexim.morganey.module.ModuleFinder
 
 import scala.annotation.tailrec
+import scala.util.Try
 
 class ReplAutocompletion(globalContext: () => List[MorganeyBinding]) extends Completer {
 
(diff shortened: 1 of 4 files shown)


keddelzz:
As of now I didn't implement autocompletion for commands, because there is only one command: `exit`. Once we have more of them, we could manage them somehow (for example: in a list) so that the REPL could autocomplete them, too.

rexim:
@keddelzz thank you very much! Sure, for the command autocompletion I've created a separate issue.

keddelzz:
Ok, nice!

keddelzz:
I have one thing to critique myself (I noticed it right now):
The autocompetion is case insensitive for bindings, but case sensitive for load-statement-paths. Do you want me to fix this before merging?

rexim:
@keddelzz if it's a quick thing, sure. Why not? :)

keddelzz:
Ok, I'll fix it right now, should be done in 5mins.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -0,0 +1,153 @@
...
+      // if nothing was typed into the repl, autocomplete with all known names
+      case _ if buffer.trim.isEmpty          => autoCompleteWith(knownVariableNames)
+      // if something was typed into the repl, autocomplete with all names starting with the input text
+      case _                                 =>
+        autoCompleteWith(knownVariableNames filter (_.toLowerCase startsWith beforeCursor.toLowerCase))
+    }
+  }
+
+  private def autocompleteLoadStatement(parts: List[String], endsWithDot: Boolean): List[String] = {
+    import MorganeyInterpreter.moduleFinder
```
rexim: Could you please provide the module finder via constructor of the ReplAutocompletion class as you did for the global context? It's actually my fault. I should've noticed that the load expression modifies the REPL state and therefore the module finder should be a part of the global context.

I'll create a separate refactoring issue for that. We will need to define the global context entity as the list of binding plus the module finder, because as we add more feature the notion of the global context will extend further.

keddelzz: Sorry, but I didn't quite get, what your asking me to do. :) Should I add a Lambda to the constructor, which gets the moduleFinder (`moduleFinder: () => ModuleFinder`, like the bindings) or as a simple parameter (`moduleFinder: ModuleFinder`) ?

rexim: Yeah, I asked to add the Lambda, right. :)

I just don't like how I accedentally created a global mutable state, that is now depended on by the ReplAutocompletion. :D

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -0,0 +1,153 @@
...
+
+    def validMorganeyElement(f: File) =
+      f.isDirectory || isMorganeyModule(f)
+
+    def topLevelDefinitions() =
+      moduleFinder.paths.filter(validMorganeyElement)
+
+    // List(root-file-of-module-path, module-file or directory)
+    def findAllModulesIn(path: String): List[(File, File)] =
+      Try(
```
ForNeVeR, line 56: Why this `Try` here? Maybe we'll want to print an error in case it happens? What kind of errors are we expecting here? If that's the convention for iterating the `paths` in `moduleFinder`, I recommend to revise it (maybe in scope of another issue).

Maybe create a new method in `MethodFinder` that will handle IO errors, you know.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
keddelzz: Ok, I got you. I'll fix this immediately! :)

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -0,0 +1,153 @@
...
+      case (xs, true)        => everythingIn(xs)
+      // load |
+      case (Nil, false)      => topLevelDefinitions().map(stripExtensionIfModuleFile).map(_.getName)
+      // load math.ari|
+      case (xs :+ x, false)  => everythingIn(xs, s => s.toLowerCase startsWith x.toLowerCase)
+    }
+  }
+
+  private object LoadStatement {
+    private val loadStatement = "load (.*)".r
```
rexim, line 95: What if some day we will change the syntax of the load expressions? Looks we will have to modify two places: `LambdaParser` and this. Is it possible to reuse `LambdaParser` for REPL autocompletion?

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -0,0 +1,153 @@
...
+      case _                           => false
+    }
+
+    lastNameInLine(line)
+      .filter(keepNamesAndComplexTerms)
+      .map(ln => countMatchingLetters(normalize(definition), normalize(ln), 0))
+      .getOrElse(0)
+  }
+
+  private val namePattern = "[a-zA-Z][a-zA-Z0-9]*"
```
rexim, line 142: Is it possible to reuse `LambdaParser` here as well?

ForNeVeR, line 142: ~~I think that calling `.r` here will save a bit of performance (because regex will be created ahead and cached between calls).~~

ForNeVeR, line 142: Ah, well, `String` doesn't have `matches(Regex)` method. Ignore my comment, it'll not help.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
rexim: Great! Thank you very much!

ForNeVeR:
Overall: awesome work. I have no strong objections against merging this, thank you!

rexim:
I have the same opinion for the overall work. I've checked it locally and it works awesome! :)

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
keddelzz, line 56: I don't think it would be nice, if the error is printed, because the autocompletion takes place when `tab` is being pressed.

I Implemented this traversal in this class, because I need more information of the module (the base directory of its morganey-class-path entry). So it did't fit in the class ModuleFinder, because of its special implementation.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
keddelzz, line 95: That would be nice, but for now I have no exact idea, how this should work.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
rexim, line 56: Hm... So this means that we will see the suppressed error when we try to actually load the module despite autocompletion not working, right?

rexim, line 56: I mean, I enter `load so`, press tab, tab didn't work, I finish expression `load somemodule`, press enter and see the error.

So this error is the reason why tab didn't work. :)

ForNeVeR, line 56: So we're trying to ignore just IO errors here, right? I don't like that other stuff, such as `validMorganeyElement`, also falls into this `Try`. And if anything inside the `Try` fails, then user will have no autocompletion at all, but we really could just skip the invalid elements and leave the valid ones. Maybe it'll be better to move this `Try` to lower scope? Something like this maybe?

``` scala
moduleFinder.paths.toStream
  .flatmap(f => Try(processIOActions(f)))
  .filter { case (_, f) => validMorganeyElement(f) }
```

Where `processIOActions` does all the fancy filesystem stuff.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
rexim, line 95: Very well then. Let's leave it as it is then. I'll create a separate issue for that, just to keep track of such things. :)

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
ForNeVeR, line 56: @rexim, I'm not sure if the proper error reporting falls into this issue, so maybe let's create another issue for this after we'll test the autocompletion under real conditions. That's not major issue at all.

But if we're going to ignore the errors, let's do it properly! (for the benefit of all the users)

rexim, line 56: @ForNeVeR sounds good to me. Let me create it...

rexim:
Alright, LGTM. Once the build is fixed, I merge this. :)

@keddelzz thank you very much for your awesome work!

keddelzz:
Sorry about the build errors! I did miss them somehow.

rexim:
@keddelzz no worries! We all are humans. That's why we have machines that check everything for us. ;)

ForNeVeR:
> We all are humans.

While that's not completely true...

> That's why we have machines that check everything for us.

...I've also missed that build failure, sorry.

rexim:
~_~

PR #97, opened 2016-07-17
Fix loading of modules on windows

### LINKED ISSUE #94: Loading of modules on Windows
If I type

```
?> load math.arithmetic
```

into the repl, it will respond with an parse error

```
[1.10] failure: `(' expected but `Î' found

SUCC := (Î»n.(Î»f.(Î»x.(f ((n f) x)))))

         ^
```
## 

I think this is a problem with the default charset of the parser used [here](https://github.com/rexim/Morganey/blob/master/src/main/scala/me/rexim/morganey/MorganeyInterpreter.scala#L87). We might use UTF-8 as the default charset for reading files.

keddelzz:
closes #94


rexim:
LGTM! Thanks!

PR #99, opened 2016-07-18
Fix autocompletion on Windows

### LINKED ISSUE #93: Autocompletion on Windows
If

```
?> load <TAB>
```

is entered into the REPL, it autocompletes to

```
load math/     load prelude
```
## 

If

```
?> load math.ari <TAB>
```

is entered into the REPL, it autocompletes to

```
load math/arithmetic
```
## 

Don't mind the `?` instead of the `λ`, the windows terminal can't display it, correctly.

keddelzz:
Closes #93

### DIFF
--- a/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
+++ b/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
@@ -7,7 +7,6 @@ import jline.console.completer.Completer
 import me.rexim.morganey.ast.MorganeyBinding
 import me.rexim.morganey.module.ModuleFinder
 import me.rexim.morganey.interpreter.InterpreterContext
-import me.rexim.morganey.util._
 
 import scala.annotation.tailrec
 import scala.util.Try
@@ -73,17 +72,13 @@ class ReplAutocompletion(globalContext: () => InterpreterContext) extends Comple
         .map { case (root, f) => (root, stripExtensionIfModuleFile(f)) }
         .filter { case (root, f) => fileNameFilter(f.getName) }
         .map(relativize)
-        .map(fixRelativePathForWindows)
+        .map(_.replace('/', File.separatorChar))
         .map(relativeFileToLoadPath)
 
     def relativize(baseAndFile: (File, File)): String = {
       val (base, file) = baseAndFile
       base.toURI().relativize(file.toURI()).getPath()
     }
-
-    def fixRelativePathForWindows(path: String): String =
-      if (runningOnWindows) path.replace('/', File.separatorChar)
-      else path
 
     (parts, endsWithDot) match {
       // load .|
--- a/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
+++ b/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
@@ -7,6 +7,7 @@ import jline.console.completer.Completer
 import me.rexim.morganey.ast.MorganeyBinding
 import me.rexim.morganey.module.ModuleFinder
 import me.rexim.morganey.interpreter.InterpreterContext
+import me.rexim.morganey.util._
 
 import scala.annotation.tailrec
 import scala.util.Try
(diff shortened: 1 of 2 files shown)


rexim:
Wanna bring this excellent comment here, 'cause it perfectly describes the root cause of the problem:

> The `relativize`-function ([here](https://github.com/rexim/Morganey/blob/88-module-finder-global-context/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala#L77)) always returns a path separated by `/`. `File.separatorChar` returns `/` on linux and `\` on windows. The replace function used [here](https://github.com/rexim/Morganey/blob/88-module-finder-global-context/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala#L14) won't replace path-separators coming from a relative path on windows, which was constructed by the `relativize`-function.

https://github.com/rexim/Morganey/issues/93#issuecomment-232584096

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -72,13 +73,18 @@ class ReplAutocompletion(globalContext: () => InterpreterContext) extends Comple
...
+        .map(fixRelativePathForWindows)
         .map(relativeFileToLoadPath)
 
     def relativize(baseAndFile: (File, File)): String = {
       val (base, file) = baseAndFile
       base.toURI().relativize(file.toURI()).getPath()
     }
 
+    def fixRelativePathForWindows(path: String): String =
+      if (runningOnWindows) path.replace('/', File.separatorChar)
```
rexim: Why does it have to be a windows specific thing? I mean let's get rid of this hacky condition and always replace `/` with `File.separateChar`. It will always work on any operating system. On Unix it will replace `/` with `/`. On Windows it will replace `/` with `\`. On anything else it will replace `/` with something else. And such approach actually hits more operating system and requires less code. What do you think?

keddelzz: That means we should we always replace `/` with `File.separatorChar`?

rexim: Yep, why not? Even though on Unix it's a useless operation on Windows it fixes a bug.  And always replacing `/` with `File.separatorChar` doesn't require to detect the OS at all.

keddelzz: Ok! I was just checking, if I did understand you correctly. I thought it would be better, if useless operations are not done. That's why I added the detection of the OS.

rexim: Yeah, I understand you. I guess I just don't like the OS detection approach. And I completely understand that it's pretty common in the JVM world, I still don't find the OS detection based on searching a specific keyword as substring of the OS name reliable enough. And if it's possible to avoid that (even by a price of a small useless operation), I would do that.

rexim: And of course, it's not my general rule. Each specific case should be considered separately. :) Here I don't see any harm in a useless operation.

keddelzz: You are absolutely right. 

## 

Another idea:
Java on Windows can handle `/` as a file separator, maybe this is a better solution than using `File.separatorChar`. With this solution the replacing of file separators won't be necessary at all.

rexim: Hm... Probably. I would consult with Mr @ForNeVeR regarding that. 'Cause he's an ultimate Windows expert. But he's already sleeping, so he will answer tomorrow. :)

In the meantime I'm gonna merge this PR once CI is passed.

rexim:
Alright, LGTM! :+1: Thank you very much!

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
ForNeVeR: I'll take a look at the issue tonight, guys. The fix seems legitimate (and good) to me, but probably we could find another way to fix the issue.

ForNeVeR: Nah, after thorough research I've decided that the current implementation is okay and it's generally not a good idea to use forward slashes (aka `/`) on Windows.

In most of the cases it will work, but they say that forward slashes won't work with UNC paths (aka `\\machinename\C:\Path`). And that's not very nice if we couldn't load Morganey modules from an UNC path, because people are using them actually (for example, UNC paths could be used as a workaround for known long path problem on Windows).

So, we're keeping the current implementation.

keddelzz: Wow, I didn't know that you can use `java.io.File` for working with UNC paths. Thank you, @ForNeVeR, for pointing that out.

ForNeVeR: Honestly I'm not sure about that neither. But my common sense tells me that `java.io.File` uses ordinary WinAPI under the hood, so it couldn't avoid to be able to work with UNC paths.

keddelzz: Your common sense was absolutely right! I did just test it!

Issue #91, opened 2016-07-13
Repl fails on input


keddelzz:
```
λ> PLUS 2 3
term: (λ m . (λ n . (λ f . (λ x . ((m f) ((n f) x))))))
```

```
λ> a -> b -> a + b
term: a
```

The problem is that in [this line](https://github.com/rexim/Morganey/blob/e2a146b3918d5aabdc11ac35888c64037963a645/src/main/scala/me/rexim/morganey/Main.scala#L38) the parse method is used instead of parseAll.

rexim:
Don't quite understand how `parseAll` would help. The syntax definition literally says that the application should be [enclosed by parenthesis](https://github.com/rexim/Morganey/blob/e2a146b3918d5aabdc11ac35888c64037963a645/kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala#L68). You will still get an error, because the correct syntax right now is `((PLUS 2) 3)` no matter how you parse it. And fixing that is the scope of #67 IMO.

keddelzz:
`parseAll` requires the parser to look at the whole input, instead of the first correct prefix. You are absolutely right, that this is wrong syntax for an application, but I'd like to get an parser error for that.

rexim:
Ah, very well then. So after this change the REPL gets a bunch of commands per user input instead of one. Just to clarify, how are we gonna evaluate them? Sequentially one by one as if they were entered separately, right?

keddelzz:
No the repl will still get only one command per user input. But it will read the entire input, instead of accepting after the first match.

rexim:
Oh, ok. For some reason I just thought parseAll repeatedly parses input until it parses everything. Looks like I was wrong. Thanks for the explanation! :)

keddelzz:
No problem. :)

keddelzz:
Can I close this issue then? It was fixed with PR #102. :)

rexim:
Ah! Yes, of course! Btw, that's why I always put `Close #<number-of-issue>` in the description of the PR, because GitHub closes the issue automatically once the PR is merged. ;)

Forgot to do that this time, though...

keddelzz:
Nice, to know thank you! I'll add `close #issue` to PRs in the future, too!

PR #102, opened 2016-07-26
Fix bug where REPL accepts on successful prefix-match

keddelzz:
It should read the full input instead.
#91


rexim:
Cool! Now if I accidentally type something like `PLUS 1 2` the behaviour of the REPL is not that confusing anymore! Thank you very much! :)

PR #103, opened 2016-07-27
Fix small bug in README.md: syntax of abstraction in BNF

keddelzz:
``` diff
-    <function> ::= ( <lambda-symbol> . <variable> <term> )
+    <function> ::= ( <lambda-symbol> <variable> . <term> )
```


rexim:
Good eye! Thank you very much! :)

Issue #89, opened 2016-07-13
Reuse LambdaParser in REPL autocompletion


rexim:
Derived from https://github.com/rexim/Morganey/pull/86#discussion_r70655203

keddelzz:
The most obvious solution: We could extract all regex (as strings) from the Parser into a separate object. Both parser and ReplAutocompletion could use it from the object. This would allow us [this](https://github.com/rexim/Morganey/blob/master/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala#L142) regex.

rexim:
And for modules autocomplete I literally did this:

``` diff
diff --git a/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala b/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
index bd329d8..e8d42cb 100644
--- a/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
+++ b/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
@@ -4,8 +4,9 @@ import java.io.File
 import java.util.{List => Jlist}

 import jline.console.completer.Completer
-import me.rexim.morganey.ast.MorganeyBinding
+import me.rexim.morganey.ast.{MorganeyBinding, MorganeyLoading}
 import me.rexim.morganey.module.ModuleFinder
+import me.rexim.morganey.syntax.LambdaParser

 import scala.annotation.tailrec
 import scala.util.Try
@@ -92,13 +93,16 @@ class ReplAutocompletion(globalContext: () => List[MorganeyBinding],
   }

   private object LoadStatement {
-    private val loadStatement = "load (.*)".r
-
-    def unapply(line: String): Option[(List[String], Boolean)] =
-      Option(line) collect {
-        case loadStatement(modulePath) =>
-          (modulePath.split("\\.").toList, modulePath.endsWith("."))
+    def unapply(line: String): Option[(List[String], Boolean)] = {
+      val result = LambdaParser.parse(LambdaParser.loading, line)
+
+      if (result.successful) {
+        val modulePath = result.get.modulePath
+        Some((modulePath.split("\\.").toList, modulePath.endsWith(".")))
+      } else {
+        None
       }
+    }
   }

   private def matchingDefinitions(line: String, knownVariableNames: List[String]): List[String] =
```

[...]

PR #104, opened 2016-07-27
Reuse LambdaParser in ReplAutocompletion

keddelzz:
Trial-implementation of issue #89.
@rexim @ForNeVeR your comments are very much appreciated!
## 

After separation of regex-patterns into an object, the regex for identifier can be reused in the ReplAutocompletion for variable names. As for the load-statement the parser had to be changed: Parsing of the module-path after the `load` keyword is now optional. The REPL has to check, if the module-path was given. During autocompletion the parser is used to verify load-statements, thanks to this change, too. 

However, if this PR is not merged, I suggest cherry-picking the abstractions `InputSource`, the implicit class in the util-package object (named `ParserOps`), which improves the usage of the parser, and the change in the file `Computation.scala`.

### DIFF
--- a/src/main/scala/me/rexim/morganey/Main.scala
+++ b/src/main/scala/me/rexim/morganey/Main.scala
@@ -41,16 +41,19 @@ object Main extends SignalHandler {
 
   def handleLine(con: ConsoleReader)(globalContext: InterpreterContext, line: String): Option[InterpreterContext] = {
     val nodeParseResult = LambdaParser.parseWith(line, _.replCommand)
-    nodeParseResult.toOption flatMap { node =>
+
+    val evaluationResult = nodeParseResult flatMap { node =>
       val computation = evalOneNodeComputation(node)(globalContext)
-      awaitComputationResult(computation) match {
-        case Success(MorganeyEval(context, result)) =>
-          result.foreach(t => con.println(smartShowTerm(t)))
-          Some(context)
-        case Failure(e) =>
-          con.println(e.getMessage)
-          None
-      }
+      awaitComputationResult(computation)
+    }
+
+    evaluationResult match {
+      case Success(MorganeyEval(context, result)) =>
+        result.foreach(t => con.println(smartShowTerm(t)))
+        Some(context)
+      case Failure(e) =>
+        con.println(e.getMessage)
+        None
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/main/scala/me/rexim/morganey/Main.scala
```diff
@@ -54,9 +51,6 @@ object Main extends SignalHandler {
           con.println(e.getMessage)
           None
       }
-    } else {
-      con.println(nodeParseResult.toString)
-      None
```
rexim, line 59: This is actually a bug. This `else` braunch printed syntax errors in REPL. After this change all the syntax errors are suppressed in REPL.

keddelzz, line 59: You're absolutely right, I didn't notice that!

src/main/scala/me/rexim/morganey/reduction/Computation.scala
```diff
@@ -29,4 +29,7 @@ object Computation {
     override def cancel(): Unit = ()
     override def future: Future[T] = f
   }
+
+  def failed[T](message: String, exception: String => Throwable): Computation[T] =
+    fromFuture(Future.failed(exception(message)))
 }
```
rexim, line 35: Well I think I understand the intention of this. You wanted to have a shortcut for creating failed computation but still be able to customize the message and the exception separately, right? But the result feels a bit weird for me to be honest. :) It feels like taking just an exception is enough, because exception already contains the message. Even `Future.failed` takes only exception.

rexim:
Ok, I want to stay clear on the problem that I wanted to avoid by creating #89. If I change the syntax of the language I want relatively easy to see which part of the project has to be changed accordingly. If REPL autocompletion uses its own regular expressions, and the interpreter uses its own set of parsers we gonna constantly see the discrepancy between those parts of the project. I personally don't know what's the best approach to deal with such problems. But I believe that centralizing everything related to the syntax makes the discrepancy problem easier to solve.

Does this PR solve the problem? Probably. It puts syntax related stuff in one place. This is good. But this is not enough. Because, anyone at any time can create just a little local helper regex somewhere for parsing syntax stuff. And it may trigger the process of syntax decentralizing over the time, because other people may use such code as a reference for their own changes. So this is something that has to be constantly prevented on the code review.

I will create a wiki page with "Things to keep in mind while developing Morganey" and will put the link to #89 their. Hopefully, we gonna use this page to track such long term problems.

For now, I'm ok with the changes in the PR and I'm ready to merge them once my remarks are fixed.

Thanks!

rexim:
@keddelzz, thanks for the latest fixes! Unfortunately I'm falling asleep. I'm gonna review and merge everything tomorrow.

keddelzz:
No worries, there's no need to rush!

rexim:
Looks good to me! Thank you very much! :+1:

rexim:
Created a wiki page https://github.com/rexim/Morganey/wiki/Things-to-Keep-in-Mind-While-Developing-Morganey

PR #105, opened 2016-08-07
Add execution mode

### LINKED ISSUE #82: Morganey Program Execution Semantic and Side-effects
During my work on #77 I faced the problem that we don't have formalized the Morganey Program Execution Semantic and its Side-effects. Which actually makes the implementation of the Morganey Interpreter pretty weird and hard to maintain.

On the stream @keddelzz suggested to implement some sort an entry point binding. For some reason I didn't like the idea at first, but after thinking about it for a while, I think it's actually pretty good! So I've tried to more or less formalize all of that and here is my vision of the Morganey Program Execution Semantic and how the side-effects are made.

Please share your thoughts.
# Morganey Program Semantic
## Morganey Module

Morganey Module is a file that consists of only bindings or loadings. No free terms.

Example:

``` morganey
// numbers.morganey

load math.arithmetic

six := (succ 5)
ten := ((plus 5) 5)
```
## Entry Point

Entry Point is a special binding inside of a Morganey module that has a special name (let it be `main`). This special name should be bound to a function that takes one argument.

Example:

``` morganey
// main.morganey

load math.arithmetic

main := (\input . <...>)
```
## Input Stream

Input Stream, it's a lazy evaluating Church pair encoded list of Church encoded numbers. This Stream represents the Standard Input of the program. The Morganey interpreter is responsible for constructing this list on the fly. The Morganey program can access this Stream only via argument of the Entry Point.
### Load Graph

When the Morganey Interpreter is given a file with a program to execute, it starts with building a Load Graph beginning from the file. A Load Graph is graph where vertices are Morganey Modules and the edges are load relations between the modules.
### Term Compilation and Execution

The interpreter traverses the Load Graph and collects all of the available bindings from it. Then it compiles the following term

``` morganey
((\binding-var1 .
((\binding-var2 .
((\binding-var3 .
<...>
((\input . <...>) <input-stream>) // <-- Entry Point term
<...>
)
binding-value3))
binding-value2))
binding-value1)
```

and reduces it. The result of reduction is interpreted as the one of the Church encoded structures and printed on the standard output accordingly.
## REPL

REPL allows the user to input loadings, bindings and free terms. REPL keeps track of all the available bindings in the global context. When a binding or a loading is entered the REPL updates its global context with the new available bindings. Each time the user enters a free term to reduce the REPL contructs the following term

``` morganey
((\binding-var1 .
((\binding-var2 .
((\binding-var3 .
<...>
((\input . <user-entered-term>) <input-stream>)
<...>
)
binding-value3))
binding-value2))
binding-value1)
```

and reduces it. The result of reduction is interpreted as the one of the Church encoded structures and printed on the standard output accordingly. Then the REPL waits for the next command.

---

This semantic is actually pretty close to what I initially wanted from Morganey when I started it.

rexim:
Add a so called execution mode which implements semantic described in #82 

So now we have three modes:
1. REPL mode. Can be entered via `sbt run`
2. Script mode. Similar to REPL mode, but simply uses a file as an input source. Can be entered via `sbt "run <filename>"`
3. Execution mode. A mode with a new semantic. Can be entered via `sbt "run exe <filename>"`

Close #82


rexim:
@ForNeVeR @keddelzz please take a look when you get a chance. Your comments are appreciated!

ForNeVeR:
Don't forget to put the spec from #82 into the project wiki.

input.txt
```diff
@@ -0,0 +1 @@
+a
```
ForNeVeR: What's up with this file? We really want it to stay in the repository root?

rexim:
> Don't forget to put the spec from #82 into the project wiki.

This is a great idea! Let me have a separate issue for that.

src/test/scala/me/rexim/morganey/LambdaInputSpecs.scala
```diff
@@ -0,0 +1,36 @@
...
+import me.rexim.morganey.church.ChurchPairConverter._
+import me.rexim.morganey.church.ChurchNumberConverter._
+import me.rexim.morganey.helpers.TestTerms
+import org.scalatest._
+
+class LambdaInputSpecs extends FlatSpec with Matchers with TestTerms {
+  val longInput = LambdaInput("khooy".toStream)
+  val singleCharInput = LambdaInput("k".toStream)
+
+  "Lambda input" should "has zero free vars by definition" in {
```
ForNeVeR: `should "have ..."`?

src/test/scala/me/rexim/morganey/LambdaInputSpecs.scala
```diff
@@ -0,0 +1,36 @@
...
+    longInput.freeVars.isEmpty should be (true)
+  }
+
+  "Lambda input" should "lazily evaluated to a pair when substituted" in {
```
ForNeVeR: And `should "be ..."` or `should "lazily evaluate ..."` :)

input.txt
rexim: Yeah, it's sort of leftover from manual testing. It was used as an input. If you enter:

```
$ sbt "run exe main.morganey" < input.txt
```

It will print `string: c`, which is proof of concept of Morganey interaction with outside world. Maybe those `input.txt` and `main.morganey` should be part of documentation and relate #106. Let me remove them...

src/main/scala/me/rexim/morganey/interpreter/MorganeyExecutor.scala
```diff
@@ -0,0 +1,46 @@
...
+      case MorganeyLoading(None) => Success(List())
+    }
+
+  def loadModuleFromReader(reader: Reader, moduleFinder: ModuleFinder): Try[List[MorganeyBinding]] =
+    LambdaParser.parseAll(LambdaParser.module, reader) match {
+      case LambdaParser.Success(nodes, _) => {
+        sequence(nodes.map(interpretNode(_, moduleFinder))).map(_.flatten)
+      }
+      case LambdaParser.NoSuccess(message, _) => Failure(new LambdaParserException(message))
+    }
```
keddelzz, line 24: Because there is an implicit for handling parse-results more easily, this could be rewritten as

``` scala
def loadModuleFromReader(reader: Reader, moduleFinder: ModuleFinder): Try[List[MorganeyBinding]] =
  LambdaParser.parseWith(reader, _.module) map {
    nodes => sequence(nodes.map(interpretNode(_, moduleFinder))).map(_.flatten)
  }
```

src/test/scala/me/rexim/morganey/interpreter/MorganeyExecutorSpecs.scala
```diff
@@ -0,0 +1,45 @@
...
+  "Executor" should "compile a correct program to a reducible term" in {
+    val expectedProgram = lapp(lvar("main"), LambdaInput(programInput)).addBindings(programBindings)
+    MorganeyExecutor.compileProgram(programInput)(rawProgram) should be (Success(expectedProgram))
+  }
+
+  "Executor" should "fail an incorrect program" in {
+    MorganeyExecutor.compileProgram(programInput)(programBindings).isFailure should be (true)
+  }
+
+  "Executor" should "note interpret empty loading nodes and just ignore them" in {
```
ForNeVeR: `note` -> `not`?

ForNeVeR:
Aside from my comments, this implementation looks good to me. :+1:

input.txt
rexim: Done

src/test/scala/me/rexim/morganey/LambdaInputSpecs.scala
rexim: done

src/test/scala/me/rexim/morganey/LambdaInputSpecs.scala
rexim: done

src/main/scala/me/rexim/morganey/interpreter/MorganeyExecutor.scala
rexim, line 24: Interesting! Let me try this...

src/test/scala/me/rexim/morganey/interpreter/MorganeyExecutorSpecs.scala
rexim: done

keddelzz:
I'm glad that this is implemented and Morganey now has a specified semantic! LGTM :+1:

src/main/scala/me/rexim/morganey/interpreter/MorganeyExecutor.scala
rexim, line 24: Done. It requeres to use `flatMap` instead of `map` though, but overall idea works pretty well! Thanks! :)

rexim:
@ForNeVeR @keddelzz cool! Thank you very much guys!

Issue #109, opened 2016-08-14
Dependent bindings problem


rexim:
Since all of the bindings are simply wrapped around the final expression https://github.com/rexim/Morganey/wiki/Execution-Mode-Semantic#term-compilation-and-execution it's not clear how to make them dependent on each other and how should it work semantically. We need to research that.

rexim:
We can [Top Sort](https://en.wikipedia.org/wiki/Topological_sorting) them though. But it doesn't allow recursive bindings anyway.

rexim:
Do we even need recursive bindings? I mean in pure Untyped Lambda Calculus there is no any recursion. It's usually achived via the Y-combinator.

ForNeVeR:
I think that's okay to require explicit application of Y-combinator from our users.

rexim:
>  I think that's okay to require explicit application of Y-combinator from our users.

Yeah, I like this idea too. I was thinking about transforming recursive bindings into code that uses Y-combinators during the compilation, but it seems too complicated for now.

We can simply forbid any cyclic binding dependencies (which allows us top-sort them before wrapping them over the entry point) and ask user to explicitly use Y-combinators.

rexim:
We also need to make sure that this REPL too

```
λ> a := 1
λ> b := a
λ> a
number: 1
λ> b
term: a
```

rexim:
https://github.com/rexim/Morganey/blob/e7b8dcd39e6f38ef5b97d407af188edf9aac29af/kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala#L12

I just realized that this filtering of unused bindings screws up the transitive bindings!!1

rexim:
Ok, I think the problem is understood more or less well. We simply don't filter bindings in advance and toplogically sort them before wrapping over an expression. In the scope of this issue I already fixed the filtering problem. For topsort I'll create a separete ticket.

Issue #79, opened 2016-07-11
Prelude mechanism


rexim:
Derived from #73 

A module in the `std` folder that is always automatically loaded in any Morganey session

### Pieces

- [x] #348 Implement prelude mechanism for execution mode
- [x] #351 Use prelude module injection mechanism for REPL mode
- [x] #346 Update samples with respect to autoloading of std.prelude module

PR #117, opened 2016-08-17
Implement list literals

### LINKED ISSUE #66: List literals
Derived from #58 

Not sure what's the best syntax for that, but maybe square brackets are ok: `[1, 2, 3]`

keddelzz:
closes #66
## 
- Lists can be specified as ranges of characters and numbers (`[1 .. 10]`, `[1, 3 .. 10]`, `[10, 8 .. 0]`)
- List literals for arbitrary terms (`[]`, `[1]`, `[[]]`, `["foo", 'b', ['a', 'r']]`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,24 +19,34 @@ To run the REPL just enter the following in the source code directory
 
 and start typing REPL commands there. The syntax of the REPL commands is
 
-    <repl-command> := <term> | <binding> | <loading>
+    <repl-command> := <term>
+                    | <binding>
+                    | <loading>
     <term> ::= <variable>
-             | <numeric-literal>
-             | <character-literal>
-             | <string-literal>
+             | <literal>
              | <function>
              | <application>
-    <binding> ::= <variable> := <term>
-    <loading> ::= load <module-path>
+    <binding> ::= <variable> ":=" <term>
+    <loading> ::= "load" <module-path>
     <module-path> := [a-zA-Z][a-zA-Z0-9.]*
 
+    <literal> ::= <numeric-literal>
+                | <character-literal>
+                | <string-literal>
+                | <list-literal>
     <numeric-literal> ::= [0-9]+
-    <character-literal> ::= '[\u0020-\u00B0]' | '\\[\\'"bfnrt]'
+    <character-literal> ::= '[\u0020-\u00B0]'
+                          | '\\[\\'"bfnrt]'
     <string-literal> ::= <java-string-literal>
+    <list-literal> ::= "[" [ term { "," term } ] "]"
+                     | "[" <numeric-literal> [ "," <numeric-literal> ] ".." <numeric-literal> "]"
+                     | "[" <character-literal> [ "," <character-literal> ] ".." <character-literal> "]"
+
     <variable> ::= [a-zA-Z][a-zA-Z0-9]*
-    <function> ::= ( <lambda-symbol> <variable> . <term> )
-    <application> ::= ( <term> <term> )
-    <lambda-symbol> ::= λ | \
+    <function> ::= "(" <lambda-symbol> <variable> "." <term> ")"
+    <application> ::= "(" <term> <term> ")"
+    <lambda-symbol> ::= "λ"
+                      | "\"
 
 The REPL will take the entered lambda term, beta-reduce it with the
 normal order reduction strategy and output the normal form of the
--- a/kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
+++ b/kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
@@ -5,6 +5,7 @@ import me.rexim.morganey.church.{ChurchNumberConverter, ChurchPairConverter}
 import me.rexim.morganey.util._
 import me.rexim.morganey.syntax.Language._
 
+import scala.collection.immutable.NumericRange
 import scala.util.parsing.combinator._
 import scala.language.postfixOps
 
@@ -27,29 +28,53 @@ class LambdaParser extends JavaTokenParsers with ImplicitConversions {
   def variable: Parser[LambdaVar] =
     identifier.r ^^ { LambdaVar }
 
-  def numericLiteral: Parser[LambdaTerm] =
+  def validNumberLiteral: Parser[Int] =
     numberLiteral.r ^? ({
-      case IntMatcher(x) => ChurchNumberConverter.encodeNumber(x)
+      case IntMatcher(x) => x
     }, { (rawInt) => s"`$rawInt' is too big"})
 
-  def characterLiteral: Parser[LambdaTerm] = (
+  def numericLiteral: Parser[LambdaTerm] =
+    validNumberLiteral ^^ ChurchNumberConverter.encodeNumber
+
+  def validCharacterLiteral: Parser[Char] = (
     escapedCharLiteral.r ^^ { s =>
-      ChurchNumberConverter.encodeNumber(escapeSequences(s charAt 2))
+      escapeSequences(s charAt 2)
     }
-    | symbolCharLiteral.r ^^ { s =>
-      ChurchNumberConverter.encodeNumber(s charAt 1)
-    }
+    | symbolCharLiteral.r ^^ { _ charAt 1 }
   )
+
+  def characterLiteral: Parser[LambdaTerm] =
+    validCharacterLiteral ^^ { c => ChurchNumberConverter.encodeNumber(c.toInt) }
 
   def stringLiteralTerm: Parser[LambdaTerm] =
     stringLiteral ^^ { s =>
       ChurchPairConverter.encodeString(unquoteString(s))
     }
 
+  private def range[T](p: => Parser[T])(implicit ev: Integral[T]): Parser[LambdaTerm] = {
+    val q = p
+    val parser = q ~ opt(comma ~> q) ~ (dotDot ~> q)
+    parser ^^ { case start ~ next ~ exit =>
+      val step  = next.map(ev.minus(_, start)).getOrElse(ev.one)
+      val range = NumericRange.inclusive(start, exit, step).toList
+      val nums  = range map ev.toInt map ChurchNumberConverter.encodeNumber
+      ChurchPairConverter.encodeList(nums)
+    }
+  }
+
+  def listLiteral: Parser[LambdaTerm] = (
+      brackets(repsep(term, comma)) ^^ ChurchPairConverter.encodeList
+    | brackets(range(validNumberLiteral))
+    | brackets(range(validCharacterLiteral))
+  )
+
   private def lambda = lambdaLetter | lambdaSlash
 
   private def parenthesis[T](p: Parser[T]): Parser[T] =
     leftParenthesis ~> p <~ rightParenthesis
+
+  private def brackets[T](p: Parser[T]): Parser[T] =
+    leftBracket ~> p <~ rightBracket
 
   def func: Parser[LambdaFunc] =
     parenthesis((lambda ~> variable) ~ (abstractionDot ~> term)) ^^ { LambdaFunc }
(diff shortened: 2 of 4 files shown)


rexim:
Thanks! I'll take a look tomorrow. :)

README.md
```diff
@@ -19,24 +19,34 @@ To run the REPL just enter the following in the source code directory
...
+                     | "[" <character-literal> [ "," <character-literal> ] ".." <character-literal> "]"
+
     <variable> ::= [a-zA-Z][a-zA-Z0-9]*
-    <function> ::= ( <lambda-symbol> <variable> . <term> )
-    <application> ::= ( <term> <term> )
-    <lambda-symbol> ::= λ | \
+    <function> ::= "(" <lambda-symbol> <variable> "." <term> ")"
+    <application> ::= "(" <term> <term> ")"
+    <lambda-symbol> ::= "λ"
+                      | "\"
```
ForNeVeR, line 49: Not sure if that should be `"\\"`.

kernel/src/main/scala/me/rexim/morganey/syntax/Language.scala
```diff
@@ -32,10 +32,18 @@ object Language {
...
 
   val leftParenthesis    = "("
 
   val rightParenthesis   = ")"
 
+  val leftBracket        = "["
+
+  val rightBracket       = "]"
+
+  val dotDot             = ".."
```
ForNeVeR: Maybe `rangeOperator` or `range`?

ForNeVeR:
Overall this looks good to me.

rexim:
@ForNeVeR thank you very much for your comments! 

I'm a bit busy right now. Will take a look at this PR later today. I hope...

README.md
keddelzz, line 49: I think either `"\"` or `"\\"` is fine, because it's not defined how to specify escape sequences in EBNF.

ForNeVeR, line 49: Again, I'm not the biggest BNF expert, but it seems that we're escaping slashes at line 39, so we should escape them here, too.

kernel/src/main/scala/me/rexim/morganey/syntax/Language.scala
keddelzz: Thank you! `rangeOperator` or `range` are definitely better names than `dotDot`.

README.md
keddelzz, line 49: I think that we can use `"\"` regardless, we differentiate between regexes and constants syntactically: Constants are written in quotes and there is no special syntax for regexes. In line 39 a regex and in line 49 a constant is used. I think thats why one is escaped and the other is not.

ForNeVeR, line 49: Well, I'm okay with that explanation. There's no need to overcomplicate things.

rexim:
Sorry, gonna review that tomorrow...

README.md
rexim, line 49: By the way, looks like this BNF sections in README starts to bloat. I think we need to move it to docs or something. I created it when it was really-really small. And back then it probably made sense. But now we should move to a better place. I'll create an issue for that.

kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
```diff
@@ -27,38 +28,65 @@ class LambdaParser extends JavaTokenParsers with ImplicitConversions {
...
+  private def range[T](p: => Parser[T])(implicit ev: Integral[T]): Parser[LambdaTerm] = {
+    val q = p
+    val parser = q ~ opt(comma ~> q) ~ (rangeOperator ~> q)
+    parser ^^ { case start ~ next ~ exit =>
+      val step  = next.map(ev.minus(_, start)).getOrElse(ev.one)
+      val range = NumericRange.inclusive(start, exit, step).toList
+      val nums  = range map ev.toInt map ChurchNumberConverter.encodeNumber
+      ChurchPairConverter.encodeList(nums)
+    }
+  }
```
rexim, line 63: Morganey is an implementation of Untyped Lambda Calculus and there are no types indeed in it. Everything is a Lambda Term and how it's rendered is just a matter of perspective.

Some interesting examples on that:

1.

```
λ> 97
char: a
λ> 'a'
char: a
```

2.

```
λ> []
number: 0
```

3.

```
λ> (\f . I)
number: 0
```

This happens because the REPL tries to detect things in the follow order:
- char
- number
- list of chars
- list of numbers
- term

But we could detect them in a different order and the result would be different. But the point is in Morganey everything is a term no matter how it's presented.

In that case, at least for me as a Morganey programmer, it is pretty intuitive to expect that the following thing would work:

```
λ> [1 .. 'a']
[1.7] failure: string matching regex `[0-9]+' expected but `'' found

[1 .. 'a']
      ^
```

`1` and `'a'` are not literals of different types. There are no types in Morganey. Morganey is an implementation of Untyped Lambda Calculus. Those are just different ways to present the same thing - Lambda Terms. So, I'd expect `[1 .. 'a']` to be equivalent to `[1 .. 97]`.

Also, the bounds of ranges cannot be expressions for some reason!

```
λ> [(1 1) .. (2 2)]
[1.8] failure: `]' expected but `.' found

[(1 1) .. (2 2)]
       ^
λ>
```

While in Haskell it's totally possible

```
GHCi, version 7.10.3: http://www.haskell.org/ghc/  :? for help
Prelude> [1 + 1 .. 2 ^ 3]
[2,3,4,5,6,7,8]
Prelude>
```

Maybe the ranges are not so literal after all. ;)

They are clearly expressions. And what we see here is not a literal, but an expression that for some unknown for me reason is evaluated outside of the language right in the parser. It feels so out of this world. It's not a Morganey expression. It's a Haskell expression (a cutted one, though, since it's not possible put an expression as a boundary) slaped onto Morganey language. And, I'm really sorry, I don't like that at all.

---

Ok, let's try to think how the Morganey ranges (not the Haskell ones) should be implemented.

First, I think we should allow any term as the boundary. Second, evaluate range inside of Morganey rather then inside of the Parser. So, let's just transform `[x .. y]` into `[x, (SUCC x), (SUCC (SUCC x)), ..., y]` and let the reduction do all the work with evaluating all of this `SUCC`-s! `x` is not a number?  Who cares! There no types in Morganey! You put it there, you asked for this exact expression: `[x, (SUCC x), (SUCC (SUCC x)), ..., y]`. It's gonna be evaluating into something. Probably a garbage. That's the Morganey way. And it always was. Everything is a term. Syntax is just sugar to **construct** terms more quickly, not to **evaluate** them.

But the above is just a concept. It doesn't answer many questions. Such as, "How to transform `[x, y .. z]`?", "When to stop wrapping `SUCC`-s over the `x`?", "How to know what to use `SUCC` or `PREV`?" and all of that should not involve evaluating terms during the parsing.

In that case, I'd probably add the following function to the prelude:

``` morganey

range := (\left . (\step . (\right . <generate the range>)))

```

and would transform `[x0, x1 .. y]` into `(((range x0) ((minus x1) x0)) y)`. And something like that I'd call a true Morganey range. But for that we need to finally implement the Prelude mechanism #79, so the `range` is always present in any Morganey session. Also resolve the dependent bindings problems #109, since the modules are almost unusable because of it. And that is why I said that the ranges should be a separate issue, because I predicted to see something like this. I simply didn't want to say anything until there is code. Just don't like to discuss unexisting things.

---

I'm gonna merge this only under the following conditions:
- We create a separate issue for implementing ranges the Morganey way and removing all of that from the parsing process.
- We give that issue the highest priority (I'm gonna work on it and its blockers on Morganey Tsoding sessions),
- Anything like that in the future will be immediately declined.

Deal?

keddelzz, line 63: I definitely get your point!

Haskell desugars its range syntax to a function call ([Source](https://www.haskell.org/onlinereport/exps.html#arithmetic-sequences)). So I think the best approach would be to mimic their behavior (as you have proposed with the `range` function).

But i don't think it's smart to remove the desugaring of the parser immediately: It's a optimization, which the parser can perform to a Morganey-program. This optimization could only be applied, if the bounds of the range are literals. If we keep this approach, it has to be tweaked to for example allow mixed usage of numbers and characters.

## 

You don't have to merge this, I can remove the ranges from the parser and we implement it later in a more Morganey-way.

rexim, line 63: > It's a optimization, which the parser can perform to a Morganey-program. This optimization could only be applied, if the bounds of the range are literals. If we keep this approach, it has to be tweaked to for example allow mixed usage of numbers and characters.

Sounds pretty good! I was thinking about introducing something like term coercing that looks like:

``` scala
// Those are similar to stuff from Church Converters
def encode[A](a: A): LambdaTerm = ???
def decode[B](t: LambdaTerm): Option[B] = ???

def termCoerce[A, B](a: A): Option[B] = decode[B](encode[A](a))
```

so we can claim that the boundaries can only be the Scala types that are Term Coercible to number. I belive such notion can reduce the combinatorial complexity of such parsing optimization. :)

> You don't have to merge this, I can remove the ranges from the parser and we implement it later in a more Morganey-way.

I just thought, maybe it's not that critical... The code that uses these ranges is probably gonna be backward compatible with the Morganey ranges that I described above... Probably, I just overreacted. Let me just think about it for a while and I'm gonna answer here later today. Thanks for contributing to this language! :)

keddelzz, line 63: > I was thinking about introducing something like term coercing that looks like:

Don't `Liftable` and `Unliftable` provide that kind of functionality?
You'd easily be able to implement something like:

``` scala
def termCoerce[A, B](a: A)
  (implicit
    lifta: Liftable[A],
    unliftb: Unliftable[B]): Option[B] =
  unliftb(lifta(a))
```

keddelzz, line 63: And if only the target type should be specified `coerceTo[Char]` you could write something like:

``` scala
def coerceTo[B: Unliftable] = new MkCoerce[B]

class MkCoerce[B](implicit unliftb: Unliftable[B]) {
  def apply[A](a: A)(implicit lifta: Liftable[A]): Option[B] = termCoerce[A, B](a)
}
```

The scala compiler will infer the type of `A` while a concrete type for `B` can be specified manually.

keddelzz, line 63: > so we can claim that the boundaries can only be the Scala types that are Term Coercible to number. I belive such notion can reduce the combinatorial complexity of such parsing optimization. :)

Oh, I see a problem with using `Liftable` and `Unliftable`: It's not accessible in `kernel`. So my idea of using that for coercion was not that useful :)

rexim, line 63: > > I was thinking about introducing something like term coercing that looks like:
> 
> Don't `Liftable` and `Unliftable` provide that kind of functionality?
> You'd easily be able to implement something like:
> 
> ``` scala
> def termCoerce[A, B](a: A)
>   (implicit
>     lifta: Liftable[A],
>     unliftb: Unliftable[B]): Option[B] =
>   unliftb(lifta(a))
> ```

Wow! I didn't realize how same this things are! Thanks! :)

> > so we can claim that the boundaries can only be the Scala types that are Term Coercible to number. I belive such notion can reduce the combinatorial complexity of such parsing optimization. :)
> 
> Oh, I see a problem with using `Liftable` and `Unliftable`: It's not accessible in `kernel`. So my idea of using that for coercion was not that useful :)

I think the fact that we need `morganey.macro._` so often in the places where it's not accessible only hints us that we are doing something wrong. Maybe we have to move everything that is in the Church Converters to `morganey.macro._` not sure yet. Let me create an issue that.

keddelzz, line 63: > I think the fact that we need morganey.macro._ so often in the places where it's not accessible only hints us that we are doing something wrong. Maybe we have to move everything that is in the Church Converters to morganey.macro._ not sure yet. Let me create an issue that.

I think the best solution for that is be to move `Liftable` and `Unliftable` from `macros` to `kernel`.

The implementation of `ChurchNumberConverter` and `ChurchPairConverter` could be moved into the definition of the typeclasses. That would allow us to deprecate `ChurchNumberConverter` and `ChurchPairConverter` and use the typeclasses instead. That would define a much cleaner interface for encoding (lifting) and decoding (unlifting) lambda terms.

rexim, line 63: Regarding the Morganey way ranges, I don't see any backward compatibility problems between them and the proposed ones (even though, it's probably not the time to think about backward compatibility, but still). So let's merge what we have here and implement what I proposed later. We gonna simply have ranges with some limitations for a while. :)

keddelzz, line 63: Ok, deal :)

rexim:
Ok, LGTM. Thank you very much!

PR #121, opened 2016-08-19
Allow values of mixed types in range bounds

keddelzz:
Drop restriction, which only allowed to use literals of the same type as the bounds for a range.
Now the user is able to write for example `[1 .. 'a']`, `['a' .. 99]` and `[1,'a' .. 'z']`.


rexim:
@keddelzz cool! Could you please update the README? :)

rexim:
Overall, LGTM. Thank you very much for this change! :+1:

keddelzz:
> @keddelzz cool! Could you please update the README? :)

Done :)

rexim:
Thanks! Let's merge it then!

PR #122, opened 2016-08-20
Shorter syntax for lambda terms

### LINKED ISSUE #67: Shorter syntax for lambda terms
It's a bit troublesome to write and read big terms right now. (At least for me.)

I suggest to add an alternative syntax, as described in [lambda calculus - notation](https://en.wikipedia.org/wiki/Lambda_calculus#Notation) to unclutter the lambda terms a bit.

But I see a problem with the last point:

> - A sequence of abstractions is contracted: λx.λy.λz.N is abbreviated as λxyz.N

Because Morganey allows variable names longer than one letter, this abbreviation rule might has to be changed to something like:
- A sequence of abstractions is contracted: λx.λy.λz.N is abbreviated as λx.y.z.N
  So it will be at least a bit shorter.

With the alternative style there could arise ambiguity-problems with the syntax. Those have to be addressed, too.

If the lambda-terms are also rendered with the new shorter syntax, this issue also addresses #51.

keddelzz:
The parser now parses the new notation introduced in #67.  
## 

TODO:
- [x] Change parser
- [x] Update docs
- [x] Add new tests
## 

Edits:
- This PR is not blocked #123 anymore after closing of #125
- Remove `Update / Port std-lib` from the list of todos, create issue #129
## 

closes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,11 @@ and start typing REPL commands there. The syntax of the REPL commands is
     <repl-command> := <term>
                     | <binding>
                     | <loading>
-    <term> ::= <variable>
-             | <literal>
-             | <function>
+    <term> ::= <function>
              | <application>
+             | <literal>
+             | <variable>
+             | "(" <term> ")"
     <binding> ::= <variable> ":=" <term>
     <loading> ::= "load" <module-path>
     <module-path> := [a-zA-Z][a-zA-Z0-9.]*
@@ -44,8 +45,8 @@ and start typing REPL commands there. The syntax of the REPL commands is
                             | <character-literal>
 
     <variable> ::= [a-zA-Z][a-zA-Z0-9]*
-    <function> ::= "(" <lambda-symbol> <variable> "." <term> ")"
-    <application> ::= "(" <term> <term> ")"
+    <function> ::= <lambda-symbol> <variable> "." { <variable> "." } <term>
+    <application> ::= <term> <term> { <term> }
     <lambda-symbol> ::= "λ"
                       | "\"
 
--- a/kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
+++ b/kernel/src/main/scala/me/rexim/morganey/syntax/LambdaParser.scala
@@ -98,7 +98,7 @@ class LambdaParser extends JavaTokenParsers with ImplicitConversions {
     loadKeyword ~> (modulePath.r ?) ^^ { MorganeyLoading }
 
   def replCommand: Parser[MorganeyNode] =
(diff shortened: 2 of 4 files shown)


keddelzz:
Do we want to rewrite the pretty-printer for the terms, too? The current implementation shows lambda terms like the old language: for example`(\f . (\x . (f x)))` it could be shown as `\f.x.f x`, though. Instead of replacing the old with the new pretty-printer we could keep both and make it configurable (as a command line flag and a repl command).

rexim:
> Do we want to rewrite the pretty-printer for the terms, too? 

I don't mind. It makes the output more readable after all.

I just thought that one priciple it would be good to stick to (which is already violated, but I'll create a separate issue for that) is everything that is printed as a result in REPL should be easily copy-pasted back to its input. I think in a language where everything is a term that makes sense. And it makes the work in REPL quite pleasant. If the old syntax is backward compatible with the new one, it's not critical to change the pretty printing right now. We can make it later if you want.

Please keep in mind, that current "pretty-printing" is just a bunch of overrided dumb `toString`-s. To output terms in the new syntax at least `LambdaFunc.toString()` should be a bit smarter.

keddelzz:
> I just thought that one priciple it would be good to stick to is everything that is printed as a result in REPL should be easily copy-pasted back to its input. 

The new format is backwards compatible with the new one, so it should already be possible to copy and paste a term, which was typed in in the new format, but output in the old, into the repl.

keddelzz:
> it's not critical to change the pretty printing right now. We can make it later if you want.

Ok, then we'll implement this later.

keddelzz:
Should we update the standard library to use the new notation? (I guess we can handle it the same way as the pretty-printer: Update the std-lib later, for now we can use the old implementation, which is compatible with the new one)

rexim:
Yeah, I think it can be done later as well. :)

Let me create issues for all of that...

rexim:
@keddelzz is this PR still WIP? If not please remove `[WIP]` from the title.

keddelzz:
I thought it was not, but I found an error just now. 
The error is that the repl shows a parse error, if a module gets loaded, which obviously should not happen.

input:

```
λ> load math.arithmetic
```

observed behavior

```
[2.1] failure: string matching regex `[
]' expected but `P' found

PLUS := (λm.(λn.(λf.(λx.((m f) ((n f) x))))))

^
```

expected behavior: loading of the module should succeed.

keddelzz:
I fixed the bug, now this PR is ready for review :)

rexim:
@keddelzz ok, cool! Gonna review on the stream.

rexim:
LGTM, thanks!

PR #125, opened 2016-08-20
Handle empty module case 

### LINKED ISSUE #123: Empty files not working
Empty modules cause an exception when evaluated, because of an invocation of the head method on an empty lists. [source](https://github.com/rexim/Morganey/blob/master/src/main/scala/me/rexim/morganey/interpreter/MorganeyInterpreter.scala#L65).
## 

Steps to reproduce the error:
- Create a empty module in the `std/` directory
  - Alternative: Create a module only containing comments in the `std/` directory
- Fire up a repl session
- Load the newly created module
## 

observed behaviour:

```
\> load emptyTest
head of empty list
```

expected behaviour:

```
\> load emptyTest
load emptyTest
```
## 

This bug blocks the implementation of #67 (PR #122).

rexim:
Should close #123

### DIFF
--- a/src/main/scala/me/rexim/morganey/interpreter/InterpreterContext.scala
+++ b/src/main/scala/me/rexim/morganey/interpreter/InterpreterContext.scala
@@ -3,7 +3,7 @@ package me.rexim.morganey.interpreter
 import me.rexim.morganey.ast._
 import me.rexim.morganey.module.ModuleFinder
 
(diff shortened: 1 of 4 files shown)


rexim:
@keddelzz could you please take a look at this PR when you get a chance?

keddelzz:
Looks good to me! :+1:

PR #132, opened 2016-08-21
Enclose chars and string in quotes

### LINKED ISSUE #126: Enclose chars in single quotes and strings in double quotes
Derived from https://github.com/rexim/Morganey/pull/122#issuecomment-241213221

>  Everything that is printed as a result in REPL should be easily copy-pasted back to its input.

For that enclose chars in single quotes and strings in double quotes here: https://github.com/rexim/Morganey/blob/e7b8dcd39e6f38ef5b97d407af188edf9aac29af/src/main/scala/me/rexim/morganey/ReplHelper.scala

rexim:
Close #126


keddelzz:
It might be nicer to implement detection of lambda terms using `Unliftable`. But this possibly is not in the scope of this issue.

Otherwise, LGTM :)

Issue #136, opened 2016-08-22
Top sort bindings before wrapping them over an expression


rexim:
Extracted from https://github.com/rexim/Morganey/issues/109#issuecomment-241452378

Issue #107, opened 2016-08-14
Switch the wiki platform


rexim:
### Final Decision

Local `doc/` folder in the src code repo
### Scope

Take `*.md`'s from wiki and put them to the `doc/` folder in the source code repo preserving the history. `git subtree` and stuff.

---

I don't like that GitHub's wiki is disconnected from the issue tracker (I cannot close issues from commit message to wiki repo). Let's search for alternatives. For example, https://readthedocs.org/ But I'm not sure if ReadTheDocs is a good place for Wiki stuff.

@ForNeVeR @keddelzz what do you guys think?

ForNeVeR:
I have been using it for another community project - [Naggum](https://github.com/codingteam/naggum). It looks nice, but requires configuration in python. Check [the config](https://github.com/codingteam/naggum/blob/develop/docs/conf.py), it's pretty big. I'd prefer some solution with lesser configuration required (and I'd even switch Naggum to it).

keddelzz:
Maybe the simplest solution is to use Morganeys repo itself. There could be a `docs/` or `wiki/` folder where all wiki-related resources and and the wiki is located. The wiki could be written in Markdown or Org-Mode, and be "build" to pdf and html using [pandoc](http://pandoc.org/) or displayed  by Github. 

Edit: That would be similar to the idea to use ReadTheDocs, though. But with pandoc the style could be specified via css.

ForNeVeR:
Well, ReadTheDocs supports custom themes too. It uses some sort of Python standard tool [Sphinx](http://www.sphinx-doc.org/en/stable/) to build the docs. And yeah, the pages will be stored in the repo, too.

I am not sure what should we do here. Maybe take a look at scaladoc?

ForNeVeR:
Probably the simplest approach would be to just put the `*.md` docs to the `doc/` directory. They will be part of main Morganey repo, and guys could then send patches to them using standard GitHub UI.

rexim:
I think I like the idea with `doc/` in the repo:
- Docs are connected with issue tracker:
  - I can close issues with commits related to the docs
  - I can refer to exact revision of a doc page
- Docs are available offline!
- We can do custom generation:
  - Our own styles
  - Our own features:
    - Issue number autodetection and generation corresponding link (yes, GitHub wiki doesn't support this simple vital shit!)
    - Any other autodetection of anything else

[...]

Issue #55, opened 2016-03-22
Functional tests


rexim:
### Pieces
- [ ] #231 Funtests Windows support
- [ ] #232 REPL funtests support
### Description

We have two modes right now:
- REPL
- Execution

The scope of this issue is to create a bunch of Morganey scripts and programs that cover all of its features, run them on CI and check if everything acts as expected.

Those test should be easily run locally on the developers machines to check stuff before pushing it.

rexim:
`echo "2 3" | sbt --error 'set showSuccess := false' run`

rexim:
http://www.scala-lang.org/api/rc2/scala/sys/process/package.html

PR #139, opened 2016-08-22
Add a couple of sample files

rexim:
Was just playing with Morganey and produced these two programs. To simply not lose them decided to put them in the repo.

They don't use std-lib because of some problems with dependent bindings #109. The problem should be resolved once #136 is implemented.


rexim:
Maybe those samples will be used in functional testing in the future #55

rexim:
Also this PR prepares the `docs` folder for #107 :)

keddelzz:
Very nice! I love how readable the new notation is. 
LGTM :+1:

ForNeVeR:
LGTM :+1:

rexim:
@keddelzz @ForNeVeR thank you guys!

Issue #113, opened 2016-08-17
Better label system for issue tracker


rexim:
I like how NixOS utilizes GitHub issue labels: https://github.com/nixos/nixpkgs/labels They are trying to emulate fields from the normal issue trackers:
- `0.kind`
- `1.severity`
- `2.status`
- `3.skill`
- `6.topic`
- `etc.`

I want to adapt such system for Morganey. In the scope of this issue I will prepare the set of fields, possible field values and instructions in which situations they have to be used. All the artifacts of this issue are going to be published in wiki.

Wanna do that after wiki platform switch. So this issue is blocked by #107

ForNeVeR:
For some reason they're missing 7.

rexim:
I think it's a secret field! For secret purposes!

keddelzz:
To be honest, I don't see the benefit of adapting the idea of NixOS' labels. But If you want to change the label system, I won't protest :)

ForNeVeR:
I think @rexim wants to experiment with GitHub issues to see if they can be made more usable by any means. I am interested in this experiment too (because we have a huge backlog of open-source projects and it's always good to know the current state of the issues there). So if @rexim wants to, I vote for that experiment.

keddelzz:
Thank you @ForNeVeR for your explanation. I didn't know that the changing of the label names will an experiment.

rexim:
Ah, @ForNeVeR knows me really well :D It's actually true. Ideally I want to come with an approach that is usable across all of the [Codingteam projects](https://github.com/codingteam/)

rexim:
Moving this issue to the `community` project: https://github.com/morganey-lang/community/issues/21 I believe it should a part of dev process guidelines.

PR #141, opened 2016-08-23
Fix unexpected behavior in unquotation macro

### LINKED ISSUE #127: Unexpected behavior in unquotation macro
Unquotation of pattern with single variable doesn't behave as expected.

Steps to reproduce the result:
- `sbt console`
- `import me.rexim.morganey.meta._`
- `val m"\\$a.b" = m"\\c.b"`

expected behavior:

```
a: me.rexim.morganey.ast.LambdaVar = c
```

observed behavoir:

```
a: me.rexim.morganey.ast.LambdaTerm = (λ c . b)
```

Description:
If only one variable is used in the unquotation-pattern it matches the whole ast instead of the valid subtree.

keddelzz:
A pattern of one variable won't match the whole term anymore. Instead it will match the correct subtree. 
closes #127

### DIFF
--- a/macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
+++ b/macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
@@ -74,16 +74,6 @@ private[meta] class QuotationMacro(val c: Context) {
     }
   }
 
-  private def hasSubpatterns(): Boolean = {
-    val patterns = c.internal.subpatterns(args.head).get
-    args.indices exists { idx =>
-      patterns(idx) match {
-        case pq"$_: $tpt" => true
-        case pq"$x @ _"   => false
-      }
-    }
-  }
-
   private def unquoteArg(i: Int): Tree = {
     val x = TermName(s"x$i")
     val subpattern = c.internal.subpatterns(args.head).get.apply(i)
@@ -134,23 +124,16 @@ private[meta] class QuotationMacro(val c: Context) {
         (q"_root_.scala.Some((..$ys))", q"_root_.scala.None")
     }
 
-    if (!hasSubpatterns() && ps.size == 1) {
-      q"""
-        new {
-          def unapply(input: $LambdaTermTpe) = Option(input)
-        }.unapply(..$args)
-      """
-    } else {
-      q"""
-        new {
-          def unapply(input: $LambdaTermTpe) = input match {
-            case $term => $ifp
-            case _     => $elp
-          }
-        }.unapply(..$args)
-      """
-    }
-
+    val identName = TermName(c.freshName())
+    q"""
+      new {
+        val $identName = true
+        def unapply(input: $LambdaTermTpe) = input match {
+          case $term if $identName => $ifp
+          case _                   => $elp
+        }
+      }.unapply(..$args)
+    """
   }
 
   private def expand() = transform(parse(buildProgram()))
--- a/macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
+++ b/macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
@@ -124,6 +124,10 @@ private[meta] class QuotationMacro(val c: Context) {
         (q"_root_.scala.Some((..$ys))", q"_root_.scala.None")
     }
 
+    /*
+     * Workaround to avoid the warning: patterns after a variable pattern cannot match (SLS 8.1.1)
+     * (See http://alvinalexander.com/scala/scala-unreachable-code-due-to-variable-pattern-message)
+     */
     val identName = TermName(c.freshName())
     q"""
       new {
--- a/macros/src/test/scala/me/rexim/morganey/meta/InterpolatorSpec.scala
+++ b/macros/src/test/scala/me/rexim/morganey/meta/InterpolatorSpec.scala
@@ -1,6 +1,7 @@
 package me.rexim.morganey.meta
 
-import me.rexim.morganey.ast.{LambdaApp, LambdaFunc, LambdaVar}
+import me.rexim.morganey.ast._
+import me.rexim.morganey.ast.LambdaTermHelpers._
 import me.rexim.morganey.helpers.TestTerms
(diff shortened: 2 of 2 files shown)


rexim:
Cool! Thank you very much! I'm gonna review that tomorrow. My brain simply stopped working after the stream. :)

keddelzz:
No problem! Take your time :)

rexim:
I was wondering why you put tag `meta` here and was about to ask that, and then I realized that you used it as a sort of component tag. Like `repl`. You ment the `meta` component of Morganey, right? :D

I created tag `meta` long time ago to mark big tasks that are currently (or should be) split into several subtasks. But since it doesn't actually work well let's use `meta` as a project component tag. I like that.

Let me remove `meta` from other places...

P.S. linking this comment with #113

keddelzz:
> I created tag meta long time ago to mark big tasks that are currently (or should be) split into several subtasks. 

Well ok, then we had a misunderstanding. I thought you meant `meta` to be the component label. :)

rexim:
Yeah, that's a funny situation. :) But it is a component tag now, I'm ok with that. Let me name the component tags accordingly...

keddelzz:
Do you know why Appveyor has a problem with building this PR?

rexim:
> Do you know why Appveyor has a problem with building this PR?

Nope. Let me restart it.

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
```diff
@@ -134,23 +124,16 @@ private[meta] class QuotationMacro(val c: Context) {
...
-            case _     => $elp
-          }
-        }.unapply(..$args)
-      """
-    }
-
+    val identName = TermName(c.freshName())
+    q"""
+      new {
+        val $identName = true
```
ForNeVeR, line 134: Could you please explain the purpose of `$identName` here? It seems to me that it's always `true` and can be safely removed.

rexim:
> > Do you know why Appveyor has a problem with building this PR?
> 
> Nope. Let me restart it.

Welp! I tried! Gonna merge that without full CI checks then. Looks like a problem on Appveyor side. Hope it's not persistent.

macros/src/main/scala/me/rexim/morganey/meta/QuotationMacro.scala
rexim: I remove that and the tests show the warning:

```
[warn] /home/rexim/Programming/rexim/Morganey/macros/src/test/scala/me/rexim/morganey/meta/InterpolatorSpec.scala:70: patterns after a variable pattern cannot match (SLS 8.1.1)
[warn]     val m"$wholeApp" = m"b c d"
[warn]         ^
[warn] /home/rexim/Programming/rexim/Morganey/macros/src/test/scala/me/rexim/morganey/meta/InterpolatorSpec.scala:84: patterns after a variable pattern cannot match (SLS 8.1.1)
[warn]     val m"$wholeApp" = m"\\b.b"
[warn]         ^
[warn] two warnings found
```

and in the Scalal REPL:

```
scala> val m"$a" = m"\\c.d"
<console>:14: warning: patterns after a variable pattern cannot match (SLS 8.1.1)
       val m"$a" = m"\\c.d"
           ^
<console>:14: warning: unreachable code due to variable pattern 'x0' on line 14
       val m"$a" = m"\\c.d"
           ^
<console>:14: warning: unreachable code
       val m"$a" = m"\\c.d"
           ^
a: me.rexim.morganey.ast.LambdaTerm = (λ c . d)

scala> 
```

rexim: Some google results on the warning message http://alvinalexander.com/scala/scala-unreachable-code-due-to-variable-pattern-message

ForNeVeR: Unfortunately, I have no idea how to apply backticks to macro code. That's probably the reason for this non-obvious clause.

rexim: If we gonna keep it in the code base it would be good to drop a comment there.

keddelzz: > Could you please explain the purpose of `$identName` here?

Of course! 
It's a workaround for the warning noticed by @rexim.

> I remove that and the tests show the warning:

As you might notice, the error only appears on patterns, which match the whole ast.
In this case the macro expands to something like:

``` scala
({
final class $anon {
  def unapply(input: me.rexim.morganey.ast.LambdaTerm) = input match {
    case (x0 @ _) => _root_.scala.Some(x0)
    case _ => _root_.scala.None
  }
};
new $anon()
}).unapply(<unapply-selector>)
```

This causes the warning, because the first pattern matches every expression.

> Unfortunately, I have no idea how to apply backticks to macro code. That's probably the reason for this non-obvious clause.

The backticks won't fix the problem, because a variable has to be use to match the whole tree.

rexim: @keddelzz could you please add a comment in the code with brief explanation about that? So anyone who will read the code in the future won't be confused. :) Thanks!

rexim: @keddelzz something really brief. Maybe even something like

```
See http://alvinalexander.com/scala/scala-unreachable-code-due-to-variable-pattern-message
```

keddelzz: Of course! I'll add it later!

rexim: @keddelzz cool! Thank you very much! :)

ForNeVeR:
LGTM :+1:

keddelzz:
The problem with Appveyor seems to be fixed, the build is passing :)

rexim:
LGTM too! :+1: 

@keddelzz thank you for the fix!
@ForNeVeR thank you for helping with the review!

Issue #131, opened 2016-08-21
TermOutputHelper is completely disconnected from the LambdaParser


rexim:
> Try to keep things related to syntax parsing in a single place and reuse them wherever it's possible.

rexim:
This comment might be related to this issue: https://github.com/rexim/Morganey/pull/132#issuecomment-241267003

rexim:
I don't like this issue. It's vague and it's unclear what exactly can be done here. Closing it.

Issue #119, opened 2016-08-19
Make morganey.macro._ accessible everywhere


rexim:
Derived from https://github.com/rexim/Morganey/pull/117#discussion_r75500216

https://github.com/rexim/Morganey/tree/84d6bbe60a2a4131bb82fad7dcd495ace90d3122/macros

Issue #59, opened 2016-07-01
Introduce Collapsed Applications


rexim:
Derived from https://github.com/rexim/Morganey/issues/58#issuecomment-229895047

Collapsed Application should store left term, right term, and a number which represents how many times the left term was applied to the right one. This may probably help with representing the numbers. We can replace `(\f . (\x . (f (f (f (f (f x)))))))` with `(\f . (\x . [f x 5]))`. And during reduction we should treat such "collapsed" Application as a series of Applications. That may really help with big numbers.

keddelzz:
Another idea is to add a special LamdaTerm (e.g. called LambdaNumber), which simply represents an integer, but this would have the consequence, that all operations have to be detected and their effect has to be simulated in scala (similarly as already mentioned in your comment in [#58](https://github.com/rexim/Morganey/issues/58#issuecomment-229900856)).

rexim:
Keep in mind this comment https://github.com/rexim/Morganey/pull/144#issuecomment-242786169 while working on this issue

rexim:
This task is really difficult to tackle at the current state of the project. I cannot even roughly subdivide this task. Putting a research label. I need to think about this more.

PR #144, opened 2016-08-25
Smartdetection using Unliftable

keddelzz:
Reimplementation of smartdetection of lambda terms in REPL, which provides:
- reduction of duplicate code: previously the detection of terms was implementet in `kernel` (`ChurchNumberConverter` and `ChurchPairConverter`) and in `src` (`TermOutputHelper`). Now it's only implemented in `kernel` and used (with the help of the `Unliftable`-typeclass) in `src`
- easier to add detection of different terms using `Unliftable`
- detection of nested terms:

```
\> ['a', [1, 2], ['a' .. 'c'], "foo", [1, 'a'], []]
elements: ['a',[1,2],"abc","foo",[1,97],0]
```

Copy and pasting the detected term is still possible.
## 

Because of this PR beeing a proposal, no additional unit tests for nested terms were added, yet.
## 

A potential problem with this implementation might arise, if #119 and #131 get implemented, because of the dependency of `TermOutputHelper` on `meta` (`Unliftable`-Typeclass).
## 

If this PR doesn't get merged, the reimplementation showed me, that I'd like to have the detection of nested terms in the REPL.

### DIFF
--- a/src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
+++ b/src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
@@ -3,38 +3,31 @@ package me.rexim.morganey.interpreter
 import me.rexim.morganey.ast.LambdaTerm
 import me.rexim.morganey.meta._
 
-import scala.annotation.tailrec
-
 object TermOutputHelper {
 
-  def smartShowTerm(term: LambdaTerm, noPrefix: Boolean = false): String = {
-    @tailrec
-    def decode(decoders: List[Decoder[_]]): String = decoders match {
-      case hd :: tl => hd(term) match {
-        case Some(result) => if (noPrefix) result else s"${hd.prefix}: $result"
-        case None         => decode(tl)
-      }
-      case Nil =>
-        // `idDecoder` won't ever return `None`
-        val result = idDecoder(term).get
-        if (noPrefix) result else s"${idDecoder.prefix}: $result"
-    }
-    decode(decoders)
+  def smartShowTerm(term: LambdaTerm): String = {
+    val (prefix, value) = decode(term)
+    s"$prefix: $value"
   }
 
-  private val idDecoder = Decoder[LambdaTerm]("term", _.toString)
+  private def decode(term: LambdaTerm): (String, String) =
+    decoders
+      .map(_(term))
+      .find(_.isDefined)
+      .flatten
+      .getOrElse(("term", term.toString))
 
-  private val decoders: List[Decoder[_]] =
-    List(
-      Decoder[Char]                     ("char", c => s"'$c'"),
-      Decoder[Int]                      ("number", _.toString),
-      Decoder[String]                   ("string", s => s"""\"$s\""""),
-      Decoder[Seq[Int]]                 ("numbers", _.mkString("[", ",", "]")),
-      Decoder[Seq[LambdaTerm]]          ("elements", _.map(smartShowTerm(_, true)).mkString("[", ",", "]"))
+  private val decoders: Stream[Decoder[_]] =
+    Stream(
+      Decoder[Char]            ("char", c => s"'$c'"),
+      Decoder[Int]             ("number", _.toString),
+      Decoder[String]          ("string", s => s"""\"$s\""""),
+      Decoder[Seq[Int]]        ("numbers", _.mkString("[", ",", "]")),
+      Decoder[Seq[LambdaTerm]] ("elements", _.map(decode(_)._2).mkString("[", ",", "]"))
     )
 
   private case class Decoder[T](prefix: String, transform: T => String)(implicit unT: Unliftable[T]) {
-    def apply(term: LambdaTerm): Option[String] = unT.unapply(term).map(transform)
+    def apply(term: LambdaTerm): Option[(String, String)] = unT.unapply(term).map(transform).map((prefix, _))
   }
 
 }
--- a/src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
+++ b/src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
@@ -1,24 +1,36 @@
 package me.rexim.morganey.interpreter
 
 import me.rexim.morganey.ast.LambdaTerm
-import me.rexim.morganey.church.ChurchNumberConverter._
-import me.rexim.morganey.church.ChurchPairConverter._
+import me.rexim.morganey.meta._
+
+import scala.annotation.tailrec
 
 object TermOutputHelper {
 
-  def isPrintableChar(code: Int): Boolean =
-    32 <= code && code <= 176
-
   def smartShowTerm(term: LambdaTerm): String = {
-    decodeNumber(term) match {
-      case Some(number) if isPrintableChar(number) => s"char: '${number.toChar}'"
-      case Some(number) => s"number: $number"
-      case None => decodeListOfNumbers(term) match {
-        case Some(numbers) if numbers.forall(isPrintableChar) =>
-          s"""string: \"${numbers.map(_.toChar).mkString}\""""
-        case Some(numbers) => s"numbers: [${numbers.mkString(",")}]"
-        case None => s"term: $term"
+    @tailrec
+    def decode(decoders: List[Decoder[_]]): String = decoders match {
+      case hd :: tl => hd(term) match {
+        case Some(result) => result
+        case None         => decode(tl)
       }
+      case Nil => idDecoder(term).get // `idDecoder` won't ever return `None`
     }
+    decode(decoders)
   }
+
+  private implicit val idDecoder = Decoder[LambdaTerm](t => s"term: $t")
+
+  private val decoders: List[Decoder[_]] =
+    List(
+      Decoder[Char]     (c  => s"char: '$c'"),
+      Decoder[Int]      (n  => s"number: $n"),
+      Decoder[String]   (s  => s"""string: \"$s\""""),
+      Decoder[Seq[Int]] (xs => s"numbers: ${xs.mkString("[", ",", "]")}")
+    )
+
+  private case class Decoder[T](transform: T => String)(implicit unT: Unliftable[T]) {
+    def apply(term: LambdaTerm): Option[String] = unT.unapply(term).map(transform)
+  }
+
 }
--- a/src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
+++ b/src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
@@ -7,29 +7,34 @@ import scala.annotation.tailrec
 
 object TermOutputHelper {
 
-  def smartShowTerm(term: LambdaTerm): String = {
+  def smartShowTerm(term: LambdaTerm, noPrefix: Boolean = false): String = {
     @tailrec
     def decode(decoders: List[Decoder[_]]): String = decoders match {
       case hd :: tl => hd(term) match {
-        case Some(result) => result
+        case Some(result) => if (noPrefix) result else s"${hd.prefix}: $result"
         case None         => decode(tl)
       }
-      case Nil => idDecoder(term).get // `idDecoder` won't ever return `None`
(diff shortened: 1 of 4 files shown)


src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
```diff
@@ -1,24 +1,41 @@
...
       }
+      case Nil =>
+        // `idDecoder` won't ever return `None`
+        val result = idDecoder(term).get
+        if (noPrefix) result else s"${idDecoder.prefix}: $result"
     }
+    decode(decoders)
+  }
+
+  private implicit val idDecoder = Decoder[LambdaTerm]("term", _.toString)
```
keddelzz: I don't know why, but this instance is `implicit`. The `implicit` keyword could be removed. This might be a autocomplete-related issue with my IDE.

rexim:
Thanks for the proposal! I'm gonna review it later today hopefully before the stream. If not then definitely tomorrow.

ForNeVeR:
From the initial post:

> ```
> \> ['a', [1, 2], ['a' .. 'c'], "foo", [1, 'a'], []]
> elements: ['a',[1,2],"abc","foo",[1,97],0]
> ```

Why is `'a'` reencoded as `'a'` at the first time (`'a'` → `'a'`), and as `97` at the second time (`[1, 'a']` → `[1,97]`)? Is it right? Is there some heuristic involved?

keddelzz:
The decoding of lists containing characters and numbers falls back to decoding a lists of numbers, if one of the numbers in not a valid character. That's the correct behavior, because characters desugar to numbers anyway. Only the detection of terms shows ascii-characters differently.

ForNeVeR:
LGTM :+1:

rexim:
Looks like I'm gonna look at that tomorrow...

@ForNeVeR thanks for helping with the review!

rexim:
> Copy and pasting the detected term is still possible.

That is not true

```
λ> pair := \x.y.z.z x y
λ> pair 1 2
pair: (1, 2)
λ> (1, 2)
[1.3] failure: `)' expected but `,' found

(1, 2)
  ^
λ> 
```

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
```diff
@@ -1,24 +1,41 @@
...
-  def smartShowTerm(term: LambdaTerm): String = {
-    decodeNumber(term) match {
-      case Some(number) if isPrintableChar(number) => s"char: '${number.toChar}'"
-      case Some(number) => s"number: $number"
-      case None => decodeListOfNumbers(term) match {
-        case Some(numbers) if numbers.forall(isPrintableChar) =>
-          s"""string: \"${numbers.map(_.toChar).mkString}\""""
-        case Some(numbers) => s"numbers: [${numbers.mkString(",")}]"
-        case None => s"term: $term"
+  def smartShowTerm(term: LambdaTerm, noPrefix: Boolean = false): String = {
```
rexim: Boolean function argument always means you need two functions instead of one

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
```diff
@@ -1,24 +1,41 @@
...
+    @tailrec
+    def decode(decoders: List[Decoder[_]]): String = decoders match {
+      case hd :: tl => hd(term) match {
+        case Some(result) => if (noPrefix) result else s"${hd.prefix}: $result"
+        case None         => decode(tl)
       }
```
rexim: I believe such pattern can be implemented via lazy application of decoders and dropping the ones that didn't work.

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
```diff
@@ -1,24 +1,41 @@
...
+      case Nil =>
+        // `idDecoder` won't ever return `None`
```
rexim: Such assumption may change in the future and the compiler won't help you.

rexim:
I like the idea but implementation is not clear and overcomplicated. I'd implement it like that ~~https://gist.github.com/rexim/abafc00d2861ee337b750d403ea5ec61~~ https://github.com/rexim/Morganey/pull/144#issuecomment-242705604

Hope it helps

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
ForNeVeR: I disagree. Consider `setVisible(visible: Boolean)`. It'd be a pain sometimes to use `show` + `hide` instead of `setVisible`.

rexim: @ForNeVeR, good point. Let's say in this case you need two functions. Please see ~~https://gist.github.com/rexim/abafc00d2861ee337b750d403ea5ec61~~ https://github.com/rexim/Morganey/pull/144#issuecomment-242705604

ForNeVeR: Agreed.

rexim:
I just realized, that posting conversation related code to gist is not a good idea, 'cause the gist can be removed in the future and the conversation may lose it's meaning. And also if you have any comments regarding that chunk of code, it would be better to keep them here.

Let me post it here:

``` scala
object TermOutputHelper {

  def smartShowTerm(term: LambdaTerm): String = {
    val (prefix, value) = decode(term)
    s"$prefix: $value"
  }

  private def decode(term: LambdaTerm): (String, String) = {
    decoders
      .map(_(term))                       // apply each encoder to the term
      .dropWhile(_.isEmpty)               // get rid of false results
      .headOption                         // try to get the head
      .flatMap(identity)                  // flatten Option
      .getOrElse(("term", term.toString)) // get result or return default
  }

  private val decoders: Stream[Decoder[_]] =
    Stream(
      Decoder[Char]                     ("char", c => s"'$c'"),
      Decoder[Int]                      ("number", _.toString),
      Decoder[String]                   ("string", s => s"""\"$s\""""),
      Decoder[Seq[Int]]                 ("numbers", _.mkString("[", ",", "]")),
      Decoder[Seq[LambdaTerm]]          ("elements", _.map(decode(_)._2).mkString("[", ",", "]")),
      Decoder[(LambdaTerm, LambdaTerm)] ("pair", { case (a, b) => s"(${decode(a)._2}, ${decode(b)._2})" })
    )

  private case class Decoder[T](prefix: String, transform: T => String)(implicit unT: Unliftable[T]) {
    def apply(term: LambdaTerm): Option[(String, String)] = unT.unapply(term).map(transform).map((prefix, _))
  }
}
```

keddelzz:
> Let me post it here

Beautiful!

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
keddelzz: > > Copy and pasting the detected term is still possible.
> 
> That is not true

You are completely right, I forgot about this case.

Detection of pairs has to be dropped for this becoming true.

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
keddelzz: I considered using the lazy application, but I went with the version which is probably is faster, due to a tail-recursive implementation.

keddelzz:
Instead of

> `.flatMap(identity)                  // flatten Option`

you could use `.flatten`

> `.flatten                                  // flatten Option`

keddelzz:
Maybe `.find` in Stream is probably better than dropping invalid decoding-results.

rexim:
> you could use `.flatten`

Right! Completely forgot about this one! :) Thanks!

> Maybe `.find` in Stream is probably better than dropping invalid decoding-results.

Yeah, nice idea! That will simplify the code even more! Thanks!

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
rexim: > Detection of pairs has to be dropped for this becoming true.

Yep. Maybe we need some sugar for pairs. Let me create and issue for that and we will discuss it there.

src/main/scala/me/rexim/morganey/interpreter/TermOutputHelper.scala
rexim: Yeah, probably. But I don't think that performance is so critical right now, that we need to sacrifice readablity in favor of non-asymptotical optimizations. Furthermore, we don't show terms to often. Just once after finishing their reduction. Maybe in the future we will need that when we have more programs written in Morganey to better analyze the performance of the interpreter.

keddelzz:
> Let me post it here

Most of my code was changed or deleted by you. Should we don't merge this then (and you create a PR with your code) or should I use your proposal and change my code to yours?

ForNeVeR:
@keddelzz, I think that you were just discussing your changes with @rexim. It's your (joy|responsibility) to either accept (and check!) his changes or argue them.

Also this was a vvery funny comment:

> Most of my code was changed or deleted by you.

Among his friends, @rexim have a reputation of a programmer who _loves_ to delete the code. And it seems he's just confirmed it again!

ForNeVeR:
@keddelzz, also this was a vvery funny comment:

> Most of my code was changed or deleted by you.

Among his friends, @rexim have a reputation of programmer who *loves* to delete code. And it seems he confirmed this again!

keddelzz:
Deleting code is equally important as writing it: Without change there's no way to progress.

keddelzz:
> It's your (joy|responsibility) to either accept (and check!) his changes or argue them.

Ok :) Then I take responsibility and apply @rexim's changes with joy.

keddelzz:
What do you think @rexim, should we drop the detection of pairs for now and re-implement it in the scope of #145?

rexim:
> @keddelzz, I think that you were just discussing your changes with @rexim. It's your (joy|responsibility) to either accept (and check!) his changes or argue them.

I second that. This is exactly what happened. I was not even trying to tell you what to do! Who am I after all to do that. :) I was just telling what problems I see in the current implementation and posted my variant for making my point more clear. My variant was also far from ideal and you showed that here https://github.com/rexim/Morganey/pull/144#issuecomment-242713588 and here https://github.com/rexim/Morganey/pull/144#issuecomment-242713971

rexim:
> What do you think @rexim, should we drop the detection of pairs for now and re-implement it in the scope of #145?

Yeah, sounds good to me.

keddelzz:
I'll add some unit tests for detection of nested terms.

keddelzz:
I just tried to use

> > ['a', [1, 2], ['a' .. 'c'], "foo", [1, 'a'], []]
> elements: ['a',[1,2],"abc","foo",[1,97],0]

as a unit test. The following compile-error arises: 

> Could not write class me/rexim/morganey/interpreter/TermOutputHelperSpec because it exceeds JVM code size limits. Method me/rexim/morganey/ast/LambdaApp's code too large!

I guess the macro generates too many applications. Maybe the implementation of #59 fixes this.

rexim:
> I guess the macro generates too much applications.

Wow! Numbers are usually the source of applications. Is it possible to split these inner structures into several test cases? Like
- `['a', [1, 2]]`
- `["foo", [1, 'a']]`

etc.

If not we can create a separate issue for unit tests that depends on #59.

keddelzz:
>  Is it possible to split these inner instructions into several test cases?

I'm trying that right now.

keddelzz:
It sort of works, I just had to do those things:
- split subexpressions
- use uppercase characters ('A' = 65, 'a' = 97)
- shorter string literals

We could add an issue for adding a unit test for 

> ['a', [1, 2], ['a' .. 'c'], "foo", [1, 'a'], []]
> elements: ['a',[1,2],"abc","foo",[1,97],0]

regardless.

rexim:
LGTM :+1: Thank you very much!

PR #146, opened 2016-08-26
Pretty printer for new short notation

### LINKED ISSUE #128: Term shorter syntax pretty-printer
Derived from https://github.com/rexim/Morganey/pull/122#issuecomment-241213221

keddelzz:
closes #128

### DIFF
--- a/kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
+++ b/kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
@@ -1,5 +1,6 @@
 package me.rexim.morganey.ast
 
+import me.rexim.morganey.ast.LambdaTerm._
 import me.rexim.morganey.church.ChurchNumberConverter._
 import me.rexim.morganey.church.ChurchPairConverter._
 
@@ -7,7 +8,7 @@ import scala.annotation.tailrec
 
 object LambdaTerm {
 
-  private def nestedApplications(app: LambdaApp): (LambdaTerm, List[LambdaTerm]) = {
+  private[ast] def nestedApplications(app: LambdaApp): (LambdaTerm, List[LambdaTerm]) = {
     @tailrec
     def go(app: LambdaApp, acc: List[LambdaTerm]): (LambdaTerm, List[LambdaTerm]) =
       app match {
@@ -17,7 +18,7 @@ object LambdaTerm {
     go(app, Nil)
   }
 
-  private def nestedFunctions(func: LambdaFunc): (List[LambdaVar], LambdaTerm) = {
+  private[ast] def nestedFunctions(func: LambdaFunc): (List[LambdaVar], LambdaTerm) = {
     @tailrec
     def go(func: LambdaFunc, acc: List[LambdaVar]): (List[LambdaVar], LambdaTerm) =
       func match {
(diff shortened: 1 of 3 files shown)


rexim:
I believe we have an issue for that #128

rexim:
Thanks for the PR! I'll take a look into it tomorrow!

keddelzz:
> I believe we have an issue for that #128

Yes, sorry, forgot to mention it, adding it now!

ForNeVeR:
LGTM :+1:

rexim:
Sorry for the recent delays in the code review. Just need to sort out some IRL bullshit. :) Will try to find some time today. If not then definitely tomorrow. Tomorrow is a Morganey dedicated stream, so I will have no other choice but review the PR. :D

keddelzz:
You do not have to justify yourself. :) Just look at the code when you have time.
`IRL bullshit` (as you called it) should always have highest priority. 
So no worries!

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
```diff
@@ -3,6 +3,59 @@ package me.rexim.morganey.ast
...
+      val rest = deep match {
+        case f: LambdaFunc => s"(${prettyPrint(f)})"
+        case t             => prettyPrint(t)
+      }
+      s"$rest $nested"
+    case func: LambdaFunc =>
+      val (vars, body) = nestedFunctions(func)
+      val variables = vars.map(t => s"${prettyPrint(t)}.").mkString("")
+      s"λ$variables${prettyPrint(body)}"
+  }
```
rexim: What I see here is that you removed polymorphic `toString` method from each implementation of trait `LambdaTerm`, created big static method and performed the dynamic dispatching between classes manually. Why?

I mean you could take each branch of this match-case expression and put them into `toString`-s of corresponding classes that you are trying to match.

This method is a huge code magnet. I cannot just inherit LambdaTerm and implement the corresponding methods anymore. I have to also not forget to add another branch to this expression which is in a different place for some reason. And over the time that `prettyPrint` method may grow infinitely.

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
```diff
@@ -3,6 +3,59 @@ package me.rexim.morganey.ast
...
+object LambdaTerm {
+
+  private def nestedApplications(app: LambdaApp): (LambdaTerm, List[LambdaTerm]) = {
+    @tailrec
+    def go(app: LambdaApp, acc: List[LambdaTerm]): (LambdaTerm, List[LambdaTerm]) =
+      app match {
+        case LambdaApp(l: LambdaApp, r) => go(l, r :: acc)
+        case LambdaApp(l, r)            => (l, r :: acc)
+      }
+    go(app, Nil)
```
rexim, line 18: Do you participate in [Competitive Programming](https://en.wikipedia.org/wiki/Competitive_programming)? :) You know ACM ICPC, TopCoder and stuff. `go` is a pretty common name there for a recursive function that performs some complex solution searching.

keddelzz: No I did not participate in Competitive Programming. :)
I chose between `go`, `helper` and `loop`.

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
keddelzz: > I mean you could take each branch of this match-case expression and put them into toString-s of corresponding classes that you are trying to match.

You are absolutely right with that. First I thought the implementations of each branch had to be more intermingled, so I started implementing it as a static function. I just kept it that way.
This is how I prefer it personally.

> I have to also not forget to add another branch to this expression which is in a different place for some reason. 

The Scala compiler will give you a warning for the not exhaustive pattern match.

## 

If you prefer the polymorphic function I'll gladly split each cases and move them into the classes.

rexim: > First I thought the implementations of each branch had to be more intermingled,

Ah, yeah, I see your point. I was thinking about that too when I was filing #128. But in the end result each branch is isolated enough. Only application tries to match functions, but this doesn't look too critical for me to keep this match-case around, so let's split the cases then. :)

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
rexim: Ah, alright then. This code just  reminded me my university days. :)

rexim:
Alright, LGTM :+1: Thank you very much!

PR #148, opened 2016-08-28
Port std-lib to shorter syntax

### LINKED ISSUE #129: Port std-lib to shorter syntax
Derived from https://github.com/rexim/Morganey/pull/122#issuecomment-241257019

rexim:
Close #129


std/boolean.morganey
```diff
@@ -1,2 +1,2 @@
-true := (\a . (\b . a))
+true := \a.b.a
```
ForNeVeR: Is there a particular reason you're using `\` here and `λ` in `arithmetic.morganey`? If no, then I think it's better to be consistent among all the standard library.

ForNeVeR: Also you haven't used space after last dot in this line, although you're using them in `arighmetic.morganey`.

keddelzz:
I'm not quite sure if that is in the scope of this issue, but I suggest implementing the whole std-lib on a very high semantic level. Therefore I would like to propose to change the definition of

```
false := 0
```

to

```
false := \a.b.b
```

(And the same thing with `nil`)

## 

Otherwise, LGTM :) :+1:

std/boolean.morganey
rexim: Done

keddelzz:
This still LGTM after you latest :+1:

rexim:
Thanks everyone!

Issue #87, opened 2016-07-13
Autocompletion for REPL commands


rexim:
Derived from https://github.com/rexim/Morganey/pull/86#issuecomment-232375532

PR #160, opened 2016-08-29
Make REPL commands distinguishable from terms

### LINKED ISSUE #83: Make commands distinguishable from terms in repl
As of now, typing `exit` into the repl closes the repl-session. But the name `exit` could also be a valid identifier in Morganey.

It would be nice, if commands would be distinguishable from variables. I propose prefixing commands with a char, that's invalid as a start of an identifier (for example `:`).

keddelzz:
- Commands are now prefixed with `:`
- Split commands to map to allow implementation of #87
## 

closes #83

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,8 @@ and start entring REPL commands there. You can enter
 - `<term>` (see [Language](#language))
 - `<binding>` (see [Language](#language))
 - `<loading>` (see [Language](#language))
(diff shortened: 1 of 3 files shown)


rexim:
LGTM :+1:

ForNeVeR:
LGTM :+1:

rexim:
@keddelzz thanks for the changes!

PR #161, opened 2016-08-30
Add command for printing raw terms without detection

### LINKED ISSUE #157: Add REPL command that lets you print the raw term (without term detection)
`raw <term>` could print Morganeys representation without auto detection of numbers, characters and lists.

```
> raw [1,2,3]
```

could print

```
\x.x (\f.x.f x) (\x.x (\f.x.f (f x)) (\x.x (\f.x.f (f (f x))) \f.x.x))
```

instead of

```
[1,2,3]
```
## 

The exact name of this command is also up for debate.

keddelzz:
`:raw 1` will now print

```
\f.x.f x
```
## 

closes  #157

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ and start entring REPL commands there. You can enter
 - `<binding>` (see [Language](#language))
 - `<loading>` (see [Language](#language))
 - `:exit` - exits the REPL
+- `:raw` - prints the representation of the term
 - `:reset` - clears the current context
 
 The REPL will take the entered lambda term, beta-reduce it with the
--- a/src/main/scala/me/rexim/morganey/Commands.scala
+++ b/src/main/scala/me/rexim/morganey/Commands.scala
@@ -1,6 +1,10 @@
 package me.rexim.morganey
 
 import me.rexim.morganey.interpreter.InterpreterContext
+import me.rexim.morganey.syntax.LambdaParser
+import me.rexim.morganey.util._
+
+import scala.util.{Failure, Success}
 
 object Commands {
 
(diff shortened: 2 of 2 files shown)


rexim:
Hm... for some reason I thought the command gonna evaluate the term and then print the result as a raw term. With this implementation if you want to see the raw result of a reduced term you first have to evaluate the term without `:raw` and then copy-paste the result to `:raw`. Simply requires two actions instead of one.

@ForNeVeR @keddelzz what do you guys think? Do we need to evaluate terms in `:raw` or not?

rexim:
I guess it's simply a matter of an expected workflow. What REPL workflow do we expect with this command?

keddelzz:
> Hm... for some reason I thought the command gonna evaluate the term and then print the result as a raw term.

Then we might had a misunderstanding. I thought this command should print the term after it was parsed without term detection.

keddelzz:
The command `raw` with its current semantic, would allow a user to see exactly how Morganey did understand an expression.

rexim:
> The command `raw` with its current semantic, would allow a user to see exactly how Morganey did understand an expression.

Yeah, good point. With the current semantic the user can do more.

rexim:
Ok, I vote for the current semantic. @ForNeVeR what do you think?

Apart from that LGTM :+1:

ForNeVeR:
Yep I'm okay with the current semantic and implementation. We may have to add a command `raw_eval` that will actually execute the expression before printing that.

LGTM :+1:

rexim:
Cool, thanks everyone!

PR #162, opened 2016-08-30
Implement autocompletion for REPL commands

### LINKED ISSUE #87: Autocompletion for REPL commands

keddelzz:
closes #87

### DIFF
--- a/src/main/scala/me/rexim/morganey/Commands.scala
+++ b/src/main/scala/me/rexim/morganey/Commands.scala
@@ -6,21 +6,24 @@ object Commands {
 
   final type Command = InterpreterContext => (InterpreterContext, Option[String])
 
-  val commandPattern  = ":([a-zA-Z]+)".r
-  val commandWithArgs = ":([a-zA-Z]+) (.*)".r
+  val commandPattern  = ":([a-zA-Z]*)".r
+  val commandWithArgs = ":([a-zA-Z]*) (.*)".r
 
-  private val commands =
+  val commands =
     Map[String, String => Command](
       "reset" -> resetBindings,
       "exit" -> exitREPL
     ) withDefault unknownCommand
 
-  def unapply(line: String): Option[Command] =
-    (line match {
+  def parseCommand(line: String): Option[(String, String)] =
+    line match {
       case commandPattern(cmd)        => Some((cmd, ""))
       case commandWithArgs(cmd, args) => Some((cmd, args))
       case _                          => None
-    }).map {
+    }
+
+  def unapply(line: String): Option[Command] =
+    parseCommand(line).map {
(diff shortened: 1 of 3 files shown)


rexim:
Gonna look at this one tomorrow. Thanks!

src/main/scala/me/rexim/morganey/Commands.scala
```diff
@@ -6,21 +6,24 @@ object Commands {
 
   final type Command = InterpreterContext => (InterpreterContext, Option[String])
 
-  val commandPattern  = ":([a-zA-Z]+)".r
-  val commandWithArgs = ":([a-zA-Z]+) (.*)".r
+  val commandPattern  = ":([a-zA-Z]*)".r
+  val commandWithArgs = ":([a-zA-Z]*) (.*)".r
```
ForNeVeR, line 14: Why change this `+` to `*`? Do we allow commands with empty names?

ForNeVeR, line 14: Ok, got that — we _are_ sort of allowing empty commands: it's convenient to think about `:` as a valid command pattern. Okay then.

ForNeVeR:
LGTM :+1:

src/main/scala/me/rexim/morganey/Commands.scala
keddelzz: > Ok, got that — we are sort of allowing empty commands: it's convenient to think about : as a valid command pattern.

This also allows re-usage of code in `ReplAutocompletion.scala`.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -127,6 +130,18 @@ class ReplAutocompletion(globalContext: () => InterpreterContext) extends Comple
     }
   }
 
+  private object IsCommand {
```
rexim, line 133: Why `IsCommand` but not `IsLoadStatement`?

keddelzz: Because I needed the type alias `Command` which is defined in object `Commands`.

## 

Edit:
Because I thought i need the type alias `Command` which is defined in object `Commands`. This is why I named the object `IsCommand` instead of `Comment` to simply not mix them up.

rexim: Ah, alright. It'd be good to be consistent on that. But it is not that critical for now.

rexim:
LGTM :+1:

rexim:
Thanks everyone!

PR #163, opened 2016-09-01
Implement removing of specified bindings

### LINKED ISSUE #149: Introduce repl-command that removes a single binding from context.
Similar to #142, but only remove a single binding.
Syntax, for example: `reset <binding>`

keddelzz:
Instead of just removing a single binding from the context, this implementation allows removing all bindings matching a regex.

Example:

```
\> load prelude
\> :reset [a-z]*t[a-z]* // (bindings containing 't')
Bindings 'true' and 'mult' were removed!
\> <tab>
nil     false   plus    succ
\>
```
## 

closes #149

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ and start entring REPL commands there. You can enter
 - `<binding>` (see [Language](#language))
 - `<loading>` (see [Language](#language))
 - `:exit` - exits the REPL
-- `:raw` - prints the representation of the term
-- `:reset` - clears the current context
+- `:raw <term>` - prints the representation of the term
+- `:reset <regex>` - clears either the current context, or all bindings matching the regex
 
 The REPL will take the entered lambda term, beta-reduce it with the
 normal order reduction strategy and output the normal form of the
--- a/kernel/src/main/scala/me/rexim/morganey/util/package.scala
+++ b/kernel/src/main/scala/me/rexim/morganey/util/package.scala
@@ -1,7 +1,8 @@
 package me.rexim.morganey
 
-import java.io.{FileInputStream, InputStreamReader, Reader, File}
+import java.io.{File, FileInputStream, InputStreamReader, Reader}
 import java.nio.charset.StandardCharsets.UTF_8
+import java.util.regex.Pattern
 
 import scala.util._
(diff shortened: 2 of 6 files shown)


rexim:
@keddelzz cool thanks! I'll try to take a look later today

ForNeVeR:
LGTM :+1:

rexim:
I'm ok with this PR in particular, but I don't like where we are going with that `:reset` command in general. I'm gonna provide my concerns and possible improvements here tomorrow and we will discuss them.

rexim:
Here are my concerns:
1. The name of the command is misleading. I understand that it is completely my fault, 'cause I'm the one who picked that name and I'm willing to fix that. This command doesn't reset anything. It unbinds bindings. I suggest to call it `:unbind`.
2. After this PR `:reset` becomes quite dangerous. With an argument it removes only subset of the bindings, without arguments it removes everything! It's like `rm` would remove your entire system if you accidentally pressed enter in the middle of typing it and didn't provide any arguments. :D I suggest to get rid of the `:reset` version that doesn't get any arguments and leave only the one that is proposed in this PR. If the user wants to remove everything, they should do `:reset .*`.
3. I think that using regexs for this purpose is kind of an overkill. Do we really need such a powerful mechanism just to define a subset of names to unbind? How about [glob](https://en.wikipedia.org/wiki/Glob_%28programming%29)? Glob is much more intuitive and not worse than regex in that particular case. I mean Unix guys use glob everywhere and it works pretty well in most of the cases. Even I personally use regexs only when I do some really precise text search in a huge set of text files with `grep`. For everything else glob is just perfect for me. And from the implementation point of view it should not be too hard. Internally Morganey can map glob expression into an equivalent regular expression and use to match the bindings as usual. And I think it can be a separate issue.

Examples:

```
\> :unbind foo      // unbind foo
\> :unbind *t*      // unbind bindings containing 't'
\> :unbind *        // unbind everything
\> :unbind          // error
```

Please share your thoughts on the above. Once we come to an agreement I will merge this PR and create corresponding issues.

keddelzz:
I agree with all your points!

rexim:
Ok, LGTM :+1: Thank you very much everyone!

PR #167, opened 2016-09-02
Move LambdaTerm helpers to classes they belong to

rexim:
I like when the logic is evenly distributed between the entities and not concentrated in a single companion object.

### DIFF
--- a/kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
+++ b/kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
@@ -1,40 +1,9 @@
 package me.rexim.morganey.ast
 
-import me.rexim.morganey.ast.LambdaTerm._
 import me.rexim.morganey.church.ChurchNumberConverter._
 import me.rexim.morganey.church.ChurchPairConverter._
 
 import scala.annotation.tailrec
-
-object LambdaTerm {
-
-  private[ast] def nestedApplications(app: LambdaApp): (LambdaTerm, List[LambdaTerm]) = {
-    @tailrec
-    def go(app: LambdaApp, acc: List[LambdaTerm]): (LambdaTerm, List[LambdaTerm]) =
-      app match {
-        case LambdaApp(l: LambdaApp, r) => go(l, r :: acc)
-        case LambdaApp(l, r)            => (l, r :: acc)
-      }
-    go(app, Nil)
-  }
-
-  private[ast] def nestedFunctions(func: LambdaFunc): (List[LambdaVar], LambdaTerm) = {
-    @tailrec
-    def go(func: LambdaFunc, acc: List[LambdaVar]): (List[LambdaVar], LambdaTerm) =
-      func match {
-        case LambdaFunc(v, f: LambdaFunc) => go(f, v :: acc)
-        case LambdaFunc(v, b)             => ((v :: acc).reverse, b)
-      }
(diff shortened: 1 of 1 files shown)


kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
```diff
@@ -82,8 +51,24 @@ case class LambdaFunc(parameter: LambdaVar, body: LambdaTerm) extends LambdaTerm
 
   override val freeVars: Set[String] = body.freeVars - parameter.name
 
+  private[ast] def nestedFunctions(): (List[LambdaVar], LambdaTerm) = {
+    @tailrec
+    def go(func: LambdaFunc, acc: List[LambdaVar]): (List[LambdaVar], LambdaTerm) =
```
keddelzz, line 56: This change will throw a `MatchError` at runtime, if this method was called from an instance with the wrong type. This method is only allowed to be called on instances of `LambdaFunc`.

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
```diff
@@ -99,12 +84,22 @@ case class LambdaApp(leftTerm: LambdaTerm, rightTerm: LambdaTerm) extends Lambda
 
   override val freeVars: Set[String] = leftTerm.freeVars ++ rightTerm.freeVars
 
+  private[ast] def nestedApplications(): (LambdaTerm, List[LambdaTerm]) = {
+    @tailrec
+    def go(app: LambdaApp, acc: List[LambdaTerm]): (LambdaTerm, List[LambdaTerm]) =
+      app match {
+        case LambdaApp(l: LambdaApp, r) => go(l, r :: acc)
```
keddelzz, line 91: Same thing as with [167-comment](https://github.com/rexim/Morganey/pull/167#discussion-diff-77370000), but the "right" type here is `LambdaApp`.

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
rexim: That's why I made `nestedFunctions` a member of `LambdaFunc`. It is not possible to invoke that `go` with something that is not `LambdaFunc`. Please correct me if I'm wrong.

keddelzz: You are right, my mistake ;) I should have pulled your branch and double checked locally before complaining :)

rexim: No-no! Please complain about my code! You made me double check it, which is good. :)

kernel/src/main/scala/me/rexim/morganey/ast/LambdaTerm.scala
rexim: It is member of LambdaApp, so I guess it's the same situation as https://github.com/rexim/Morganey/pull/167#discussion_r77370535

keddelzz: > so I guess it's the same situation as

Yes :)

keddelzz:
LGTM :+1:

rexim:
@keddelzz cool! Thank you very much for the review!

PR #176, opened 2016-09-04
Handle cycled loads

### LINKED ISSUE #80: Handle cycled loads
Derived #73 

Don't know what's the best way to handle cycled loads, but we may try to simply ignore load expression when the same module is tried to be loaded more than once.

rexim:
Allow cycled module loading. Simply ignore modules that were already loaded. Also, finally got rid of old Morganey semantic completely.
## 

Close #80


ForNeVeR:
LGTM :+1:

keddelzz:
The implementation of `MorganeyRepl.evalNode` is quite long, maybe private helper functions will increase readability. Otherwise LGTM :+1:

rexim:
> The implementation of `MorganeyRepl.evalNode` is quite long, maybe private helper functions will increase readability.

Interesting thing is that the entire implementation consists of a single match-case which is actually a single object dynamic dispatching. Which hints that we are dealing with a hidden entity here.

A good solution would be to introduce a typeclass `ReplEvaluable` and implement the instances for each of the case-branch.

I'll create a separate issue for that.

keddelzz:
I think a typeclass for the evaluation is a bit overkill. Why not split the cases to private functions?

rexim:
> I think a typeclass for the evaluation is a bit overkill. Why not split the cases to private functions?

Because we already have lots of hidden entities that are "REPL Evaluable". But for some reason they sit in a couple of singleton objects in a form of vague functions:
- [Commands](https://github.com/rexim/Morganey/blob/8f9dbbc4aa558a3cb3e59c178bfcca76aa4fea6a/src/main/scala/me/rexim/morganey/Commands.scala)
- [MorganeyRepl](https://github.com/rexim/Morganey/blob/48b3b03022bcf1d9ab7ed6d60e98a95724c6de7d/src/main/scala/me/rexim/morganey/interpreter/MorganeyRepl.scala)

And I'm pretty sure we gonna add more of such entities.

rexim:
Cool! Thank you guys for the review!

PR #193, opened 2016-09-12
Fix pretty-printer not handling precedence correctly

keddelzz:
closes #188

PR #194, opened 2016-09-12
Use Scala 2.11.8 on Travis CI

rexim:
### Description

I compared Morganey with my [other scala project](https://github.com/tsoding/Issuestant/blob/master/.travis.yml) that I setup recently. And everything works fine there. So I tried to make Morganey's travis config closer to the Issuestant's one. Looks like it helped.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: scala
+jdk: oraclejdk7
+sudo: false
 scala:
   - 2.11.8
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: scala
-jdk: oraclejdk7
-sudo: false
 scala:
-  - 2.11.7
+  - 2.11.8
+script:
+  - sbt retest
 notifications:
   email:
     - reximkut@gmail.com


keddelzz:
LGTM :+1:

rexim:
Even though the build for #193 is fixed without this PR, I think it still better to keep the scala version in `.travis.yml` and `build.sbt` in sync.

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: scala
-jdk: oraclejdk7
-sudo: false
```
rexim, line 3: Also, I'm not sure if it was a good idea to get rid of these two. Let me get them back and see if it doesn't break anything.

rexim:
Hm... looks like Mr @ForNeVeR returned back to his cave. Alright! Not waiting for him then!

ForNeVeR:
> back to his cave

I told you, it's not the cave but the [kathisma](https://en.wikipedia.org/wiki/Kathisma#Monastic_cell) where I'm praying the Machine God, executing important rituals, and sometimes — for sure! — degustating the best canaries of machine oil.

And, for that matter, I approve this PR. LGTM 👍

rexim:
Thank you very much @ForNeVeR! :laughing:

Issue #201, opened 2016-09-18
Extend parser to support non-literal range literals


keddelzz:
subtask of #120

Issue #120, opened 2016-08-19
Morganey way ranges


rexim:
Derived from https://github.com/rexim/Morganey/pull/117#discussion_r75447348

keddelzz:
This issue was split into:
- [x] #201: Extend parser to support non-literal range literals
- [x] #202: Implement functions to construct ranges

## 

Can this issue be closed, or does #204 also count as a sub-task of #120?

rexim:
@keddelzz yeah, I think #120 can be closed now. Thanks!

PR #202, opened 2016-09-18
Implement functions to construct ranges

keddelzz:
The parser still has to be extended to support creation of non-literal range literals. This falls in the scope of the issue #201.
## 

splits #120


rexim:
Thanks! Looks really awesome! I'm gonna review that tomorrow.

It'd be great to have some funtests on that #55 But the funtests infrastructure is not merged to the master yet. :) So it can be done later separately.

ForNeVeR (approved):
Awesome 👍

keddelzz:
> It'd be great to have some funtests on that #55 But the funtests infrastructure is not merged to the master yet. :) So it can be done later separately.

I created an issue for that. #204

rexim:
It feels like the implementation of the range could be much faster

```
$ time cat test1.mgn | java -jar morganey.jar 
λ> load prelude
λ> 
λ> range 1 20
numbers: [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20]
λ> 
real    0m16.993s
user    0m18.168s
sys 0m0.229s

------------------------------

$ time cat test2.mgn | java -jar morganey.jar 
λ> load prelude
λ> 
λ> range2 := Y (\rec.start.end. (le start end) (cons start (rec (succ start) end)) nil)
λ> range2 1 20
numbers: [1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20]
λ> 
real    0m0.855s
user    0m1.821s
sys 0m0.079s
```

I understand that the performance is not our primary concern, but c'mon. :) 10 times.

rexim:
@keddelzz thanks for the improvement! I'm gonna take a look after the stream.

rexim (approved):
LGTM :+1: Thanks! It's nice to see how our std-lib grows. :)

PR #209, opened 2016-09-20
Return failed computation, if command doesn't exist

### LINKED ISSUE #207: Return failed computation, if command doesn't exist
Instead of returning a successful computation, it should be returned a failed computation, if the command doesn't exist (see [here](https://github.com/morganey-lang/Morganey/blob/49e37dc9505721634058bf6adbff14338daca461/src/main/scala/me/rexim/morganey/interpreter/MorganeyRepl.scala#L19)).
## 

This issue is related to [let-morganey-rest#8](https://github.com/morganey-lang/let-morganey-rest/issues/8).

keddelzz:
closes #207
## 

It would be a much more elegant solution if failures are not expressed by `Computation.failed`, but with `ReplResult.failed`. We might want to do that in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,4 +124,4 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 [scala-sbt]: http://www.scala-sbt.org/
 [execution-semantic]: https://github.com/rexim/Morganey/wiki/Execution-Mode-Semantic
 
-<!-- ForNeVeR ate all of my waffles again -->
+<!-- keddelzz and ForNeVeR ate all of my waffles again -->
--- a/README.md
+++ b/README.md
@@ -124,4 +124,4 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 [scala-sbt]: http://www.scala-sbt.org/
 [execution-semantic]: https://github.com/rexim/Morganey/wiki/Execution-Mode-Semantic
 
-<!-- keddelzz and ForNeVeR ate all of my waffles again -->
+<!-- keddelzz ate all of my waffles again -->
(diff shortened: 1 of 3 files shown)


rexim:
lol, looks CIs are freaked out because of the repo transfer %)

rexim:
> It would be a much more elegant solution if failures are not expressed by `Computation.failed`, but with `ReplResult.failed`. We might want to do that in the future.

It would be also great to have some kind of stable well documented API that the users can rely on though the major version of Morganey. And we don't even have any releases or versioning systems! I'll create separate issues for all of that. It's time to upgrade our process a bit!

rexim:
@keddelzz could you please make a useless commit and push it just to poke our CIs? I granted them all of the required rights, so they should work. You can remove the very last line of `README.md` file. It's an HTML comment with a secret information. I left it there to be deleted later in case of such situations. Thanks!

ForNeVeR (approved):

ForNeVeR:
> And we don't even have any releases or versioning systems!

Ya know, I think it's time to have some. There's the whole ecosystem around Morganey already, and all of the guys actually need the binaries, releases and versioning.

keddelzz:
I've to admit, that now I want to really eat some waffels.

keddelzz:
Seems like Travis is fine now, but AppVeyor still doesn't like me (or my PR).

rexim:
Appveyor job is probably queued because of #211. Let's wait a little bit

keddelzz:
> Appveyor job is probably queued because of #211.

Oh well, didn't see that! My mistake!

rexim:
Alright, I know what was the problem. Unlike Travis Appveyor doesn't recognize repo transfer. So I had to to the settings and update the repo path. We need to poke the PR again. I propose to introduce some kind of CI poke counter in the secret section of README, because such situation are not rare. :)

ForNeVeR:
I propose to **not** mention me again in that counter! >_<

rexim:
Oh my gosh! I have the best idea ever! We need to create a poke badge with the poke counter! :D

rexim:
> I propose to not mention me again in that counter! >_<

:(

keddelzz:
> I propose to not mention me again in that counter! >_<

Removed your name in the last poke.

rexim (approved):
LGTM :+1: Thanks!

Issue #143, opened 2016-08-25
Shorter syntax usability problem


rexim:
After using the shorter syntax for a while I found a problem that was not visible during the design time. This problem comes from the fact that we use the same character for separating function arguments and its body and function arguments from each other. For example,

```
\x.y. x => (\x . (\y . x))
\x.y x  => (\x . (y x))
```

Maybe it's just me I dunno, but sometimes I simply forget that last dot and it completly changes the entire meaning of my expression. I want to address that. One of the possible solutions would be simply use different symbols for arguments and body separation, like

```
\x,y. x => (\x . (\y . x))
\x,y x => Syntax error
```

rexim:
But overall I like the new syntax. It really helps with constructing complex terms. :)

keddelzz:
I'm not sure if I like the proposed notation `\x,y. x`. The `,` seems more similar to a parameter separator of a c like programming language, but the current notation is more common for lambda-calculus. (See: [Wikipedia - Lambda calculus - Notation](https://en.wikipedia.org/wiki/Lambda_calculus#Notation))

Maybe it helps to think "a dot after each parameter" instead of "a dot separates parameters from function body".

rexim:
Ok, let me use it for more time and I'll get back to this discussion with the feedback later. Maybe it's really just me.

rexim:
A few more comments from my side

> The `,` seems more similar to a parameter separator of a c like programming language, but the current notation is more common for lambda-calculus.

It's not about comma, it's about using a different symbol for a different purpose. Just want to be clear on that. We can still use dot, but the last dot replace with arrow `\x.y -> x`. It doesn't matter.

> Maybe it helps to think "a dot after each parameter" instead of "a dot separates parameters from function body".

Notation on the wiki page has fundamentally different property which I actually like:

```
λx.λy.λz.N -> (λx. (λy . (λz . N)))
λx.λy.λz N -> Syntax error
```

You just can't write it incorrectly.

And again, I'm gonna use the current syntax a bit longer. Maybe later I'll have a different view on that. Thanks!

rexim:
@keddelzz, I think you was right. Not experiencing this problem anymore. Thanks!

PR #213, opened 2016-09-21
Add more std-lib functions

keddelzz:


keddelzz:
@rexim @ForNeVeR, i hope you're okay with me creating the label `component: std-lib`.

std/list.mgn
```diff
@@ -25,6 +27,22 @@ rangeWithNext := \start.next.end. rangeWithStep start (minus next start) end
  */
 rangeWithStep := Y (\rec.start.step.end. (le start end) (cons start (rec (plus start step) step end)) nil)
 
+/* Maps function `f` over the elements of list `xs`
+ */
+map := \xs.f. Y (\mapRec.xs.f. (isnil xs) [] (cons (f (head xs)) (mapRec (tail xs) f))) xs f
```
rexim: I'd go with the same approach as Haskell: http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#v:map

The map takes the function first. I think it's better because the following partial applications are more common:

``` morganey
succAll := map succ
succAll [1 .. 10]
```

std/list.mgn
```diff
@@ -25,6 +27,22 @@ rangeWithNext := \start.next.end. rangeWithStep start (minus next start) end
...
+
+/* Creates a list containing `n` times `x`
+ */
+replicate := \x.n. Y (\replicateRec.n.x.i. (lt i n) (cons x (replicateRec n x (succ i))) []) n x 0
```
rexim: I'd take the number first for the same reason as for `map`. Binding the number of repeatitions is more common then binding `x`. Haskell agrees with me http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#v:replicate

std/list.mgn
```diff
@@ -25,6 +27,22 @@ rangeWithNext := \start.next.end. rangeWithStep start (minus next start) end
...
+
+foldl := \xs.zero.op. Y (\foldlRec.xs.acc.op. (isnil xs) acc (foldlRec (tail xs) (op acc (head xs)) op)) xs zero op
```
rexim: `\op.zero.xs.`. The same reason. It's more common to bind operation and the base element. http://hackage.haskell.org/package/base-4.9.0.0/docs/Prelude.html#v:foldl

rexim: Also `op` is already is the scope of inner function. No need to thread it through the entire iteration:

`\op.zero.xs. Y (\foldlRec.acc.xs. ???...`

std/list.mgn
```diff
@@ -25,6 +27,22 @@ rangeWithNext := \start.next.end. rangeWithStep start (minus next start) end
...
+
+foldr := \xs.zero.op. foldl (reverse xs) zero (\a.b. op b a)
```
rexim: `\op.zero.xs.`

rexim (changes requested):

keddelzz:
I do like Scala's parameter ordering better. In my opinion it's better to remember.

rexim:
@keddelzz what does make them easier to remember in your opinion?

keddelzz:
For me it's easier to remember, because it's the way I think about the operations.
- The function `map` maps the list `xs` with function `f`

```
map := \xs.f.  ...
```

## 
- Creates a list containing `n` times `x`

```
replicate := \x.n.  ...
```

## 
- The function `fold` folds together a list `xs` using a `zero` element if the list is empty. The function `op` is used to combine the accumulator and an element.

```
foldl := \xs.zero.op.  ...
```

## 

I see the convenience, which is achieved by using Haskell's parameter ordering, but then I'll have to learn how to read the functions instead :)

rexim:
Well, sometimes we have to change our habbits ;) _cough_ #143 _cough_

---

I mean I see where it comes from. In Scala all of these operations are methods of some objects. So technically you just can't  put `xs` not first. But there are no objects in Morganey...

@ForNeVeR @kuchenkruste @06needhamt what do you guys think? Which argument order is better for Morganey? Haskell or Scala?

rexim:
cc @Minoru

Minoru:
Disclaimer: I know Haskell better than I know Scala. I also like Haskell better than Scala.

I agree with @rexim's reasoning: since Morganey supports partial application, the arguments should be ordered such that the most common partial applications are convenient to write. (For everything else, there's `flip`.)

keddelzz:
Ok, let's wait on the answer of @ForNeVeR @kuchenkruste and @06needhamt.

ForNeVeR:
We have the current argument order in Scala because of its object-oriented origin: `map` in Scala is the **method** of `Seq` trait that takes one argument. It's uncommon (or even syntactically impossible) to use method without the containing class in Scala, and that's the exact reason it's designed like that.

After all, `collection.map(function)` is perfectly okay and feels logical for an object-oriented world, but I doubt `function.map(collection)` would be the same. _(You could probably infer some additional understanding from connecting `map` method with its typeclass origin, but let's stay aside from that for a moment.)_

On the other hand, Morganey, like Haskell, is not an object-oriented language. `map` is the first-class citizen in those languages. That's the reason it takes the function argument first: in a functional world it just makes more sense.

Summary: I agree with @rexim and vote for Haskell-style order of the arguments.

keddelzz:
Ok, I think we waited enough for @kuchenkruste and @06needhamt to answer. I'll change the parameter ordering to Haskell's for convenient partial application.

rexim:
@keddelzz thanks! I'll try to review the changes  before the stream. If not then tomorrow.

rexim (approved):
LGTM :+1: Thank you very much for the changes!

PR #220, opened 2016-09-26
Create tiny-jar for std-lib

### LINKED ISSUE #218: Pack std-lib as a separate jar for tiny jars version of Morganey
Derived from the discussion on the stream.

Just keep the std-lib in a separate jar `morganey-stdlib.jar`

keddelzz:
closes #218

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,7 +51,7 @@ packStdLib := {
     base.toURI.relativize(f.toURI).getPath
 
   val stdLibJar = filesIn(rootDir / "std").map(f => f -> relativeTo(rootDir, f))
-  val stdLibJarFile = targetDir / s"${name.value}_2.11-${version.value}-stdlib.jar"
+  val stdLibJarFile = targetDir / s"${name.value}-stdlib_2.11-${version.value}.jar"
 
   IO.jar(stdLibJar, stdLibJarFile, new java.util.jar.Manifest())
   stdLibJarFile
--- a/build.sbt
+++ b/build.sbt
@@ -63,6 +63,7 @@ build := {
   val (_, coreFile) = packagedArtifact.in(Compile, packageBin).value
   val (_, macroFile) = packagedArtifact.in(macros).in(Compile, packageBin).value
   val (_, kernelFile) = packagedArtifact.in(kernel).in(Compile, packageBin).value
+  val _ = packStdLib.value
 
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -32,12 +32,39 @@ lazy val dependencySettings =
...
+  stdLibJarFile
+}
+
 lazy val build = TaskKey[Unit]("build", "Builds morganey and copies all generated files into the target folder.")
 
 build := {
   val (_, coreFile) = packagedArtifact.in(Compile, packageBin).value
   val (_, macroFile) = packagedArtifact.in(macros).in(Compile, packageBin).value
   val (_, kernelFile) = packagedArtifact.in(kernel).in(Compile, packageBin).value
+  val _ = packStdLib.value
```
keddelzz, line 67: The assignment in this statement can safely be replaced by the method call (without assignment).

``` scala
packStdLib.value
```

ForNeVeR: I'm not sure, but could you maybe try to move this statement to `dependsOn` clause? Like `dependsOn(packStdLib, macros % dependencySettings, ...)`?

rexim (approved):
LGTM :+1:

rexim:
Waiting on the @ForNeVeR's resolution

ForNeVeR (changes requested):
I'd like @keddelzz to check my suggestion. If it won't work — then LGTM.

build.sbt
keddelzz: This is not possible. `packStdLib` is of type `TaskKey[File]` and `dependsOn` requires a value of type `Project`. However dependencies on tasks can be expressed differently:

For example:

``` scala
myTask <<= (aTask, bTask) map { (a: A, b: B) =>
  ... do something with a, b and generate a result ...
}
```

Example taken from SBT-documentation [here](http://www.scala-sbt.org/0.12.1/docs/Detailed-Topics/TaskInputs.html).

Should I rewrite both tasks `build` and `packStdLib` in terms of this way to specify dependencies?

ForNeVeR: Okay then, I approve the current implementation. Forgive the old man for confusing you.

ForNeVeR (approved):

rexim:
@keddelzz thanks for the changes!

Issue #231, opened 2016-09-28
Funtests Windows support


rexim:
Split from #55 

The problem is if you just uncomment [appveoyr.yml:10](https://github.com/morganey-lang/Morganey/blob/38defbbbb79aacb0f482967e3894acb2c46279f8/appveyor.yml#L10) the funtests hang on Appveyor

PR #235, opened 2016-10-01
#234: fix console redirection issues on Windows

### LINKED ISSUE #214: Morganey Hangs when running from inside intellij

### LINKED ISSUE #234: Morganey hangs when input is redirected on Windows
_(Extracted from #214.)_

This completely hangs on my machine (Windows 10 x64) when ran from `cmd.exe` shell:

``` console
$ sbt assembly
$ echo :exit | java -Dfile.encoding=windows-1251 -jar target\scala-2.11\morganey.jar
?>
```

I think that it should exit immediately after the start, but it just keeping showing me a `?>` prompt and not accepting any input at all. It eats 100% of one CPU core while doing that.

@rexim claims the same works okay on Linux.

ForNeVeR:
### Description

Fixes #234; fixes #214. The core problem lies in the JLine implementation. While debugging the issue, I found that [this line](https://github.com/jline/jline2/blob/72641325da0d15efd2555d377800d20840bad7d6/src/main/java/jline/WindowsTerminal.java#L164) was constantly throwing and swalloging `ReadConsoleInputW` exceptions. `ReadConsoleInputW` is a Windows API function, and its binding is defined somewhere in [jansi-native](https://github.com/fusesource/jansi-native) that's used by JLine. This function (and Windows Console API in general) cannot be called if the process' input or output is redirected (i.e. it have no its own console window to write to or to read from).

So, JLine tries to call the function and it cannot. That's the reason it hangs completely. So, I've decided to implement simple detection of case "we're running on Windows and our input is redirected". I've copied the Windows detection from jline [`WindowsTerminal.java`](https://github.com/jline/jline2/blob/f9b8398fa92b7de099ab9dd27b83b86239cc0516/src/main/java/jline/TerminalFactory.java#L97-L98), and I detect whether input is redirected by checking if `System.console()` exists. I believe these heuristics are decent enough to do the right thing™.
### Checklist
- [ ] ~~Update docs~~ — not necessary
- [ ] ~~Add/Update Unit Tests~~ — will be updated in scope of #231

### DIFF
--- a/src/main/scala/me/rexim/morganey/Main.scala
+++ b/src/main/scala/me/rexim/morganey/Main.scala
@@ -50,8 +50,7 @@ object Main extends SignalHandler {
 
     val running = true
     var globalContext = context
-    val con = new ConsoleReader()
-    con.setPrompt("λ> ")
+    val con = initializeConsoleReader()
     con.addCompleter(new TerminalReplAutocompletion(() => globalContext))
 
     def line() = Option(con.readLine()).map(_.trim)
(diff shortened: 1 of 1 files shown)


rexim (approved):

rexim:
@ForNeVeR also, have you checked this fix in the context of #214? Does it fix the hanging in Intellij IDEA?

ForNeVeR:
@rexim yes it fixes #214.

rexim:
Not waiting for @keddelzz, because it would be good to merge this ASAP to let @06needhamt check the fix.

@keddelzz, please feel free to provide your review at any time after the merge. All of your remarks are going to be converted to separate issues. Thanks!

@ForNeVeR, thanks for the fix!

keddelzz (commented):
LGTM :+1:

06needhamt:
Thanks Will check this now

06needhamt:
Thanks guys everything is working fine now

PR #262, opened 2016-10-16
Implement take function for std-lib

rexim:
`take` function takes `n` first elements from list `xs`


keddelzz (approved):
LGTM 👍

rexim:
@keddelzz thanks for the review!

PR #287, opened 2016-10-30
Autocomplete unbind

### LINKED ISSUE #285: Autocomplete bindings in :unbind command

keddelzz:
closes #285

### DIFF
--- a/src/main/scala/me/rexim/morganey/Commands.scala
+++ b/src/main/scala/me/rexim/morganey/Commands.scala
@@ -9,12 +9,17 @@ import scala.util.{Failure, Success}
 
 object Commands {
 
-  final type Command = ReplContext => Computation[ReplResult[String]]
+  type CommandTask = ReplContext => Computation[ReplResult[String]]
 
-  private def command(f: ReplContext => ReplResult[String]): Command =
+  case class Command(
+    name: String,
+    task: String => ReplContext => ReplResult[String],
+    hasTermArgument: Boolean)
+
+  private def command(f: ReplContext => ReplResult[String]): CommandTask =
     { context: ReplContext => Computation(f(context)) }
 
-  private def unknownCommand(command: String): Command =
+  private def unknownCommand(command: String): CommandTask =
     { context: ReplContext => Computation.failed(new RuntimeException(s"Unknown command '$command'!")) }
 
   val commandPattern  = ":([a-zA-Z]*)".r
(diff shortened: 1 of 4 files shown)


rexim:
@keddelzz that is so cool! Thanks! :) But I'm gonna review that later. Falling asleep already -_-

keddelzz:
>   But I'm gonna review that later.

No problem. :)

src/main/scala/me/rexim/morganey/Commands.scala
```diff
@@ -9,24 +9,31 @@ import scala.util.{Failure, Success}
...
 object Commands {
 
-  final type Command = ReplContext => Computation[ReplResult[String]]
+  type CommandTask = ReplContext => Computation[ReplResult[String]]
 
-  private def command(f: ReplContext => ReplResult[String]): Command =
+  case class Command(
+    name: String,
+    task: String => ReplContext => ReplResult[String],
+    hasTermArgument: Boolean)
```
rexim: Can we get rid of that? I mean it's no better than having a function with a boolean argument. Why can't it be something like

``` scala
trait Command
case class StringCommand extends Command
case class TermCommand extends Command
```

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -127,18 +130,34 @@ object ReplAutocompletion {
...
-  private object IsCommand {
+  private object SimpleCommand {
 
     def unapply(line: String): Option[List[String]] = {
       val potentialCommand = parseCommand(line).map(_._1)
       val matchingCommands = potentialCommand map { p =>
-        commands.keySet filter (_ startsWith p)
+        commands.values collect {
+          case c if c.name startsWith p =>
+            if (c.hasTermArgument) c.name + " " else c.name
```
rexim: I don't understand this code. Could you please elaborate on what's going on here? Why do we care about term arguments when we parse a simple command?

rexim (changes requested):

ForNeVeR (approved):
I have no objections except from what @rexim said.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
keddelzz: My goal with this code, was to autocomplete commands that have a term as an argument to ":command " (notice the space at the end). The space was, as far as I thought, necessary that the autocompletion for commands with arguments (`CommandWithArg`) can distinguish the command name and the arguments.

However, I was wrong. The space isn't needed.

I should have tested that earlier.

rexim:
@keddelzz thanks for the update! I'm gonna review it a bit later today.

src/main/scala/me/rexim/morganey/ReplAutocompletion.scala
```diff
@@ -127,18 +130,39 @@ object ReplAutocompletion {
...
+      cmd match {
+        case StringCommand(_, _) => false
+        case TermCommand(_, _)   => true
+      }
+
+    def unapply(line: String): Option[(String, String)] = {
+      val potentialCommand = parseCommand(line)
+      potentialCommand flatMap { case (p, arg) =>
+        commands.values find { cmd =>
+          cmd.name == p && hasTermArgument(cmd)
```
rexim: Could you please inline `hasTermArgument` here like

``` scala
commands.values find {
  case TermCommand(name, _) => name == p
  case _ => false
}
```

rexim (changes requested):

rexim (approved):

rexim:
@keddelzz thank you very much for the changes!

Issue #84, opened 2016-07-11
Selectable reduction strategy


keddelzz:
I'd like to have the possibility to choose which reduction strategy is used to evaluate a script or a term in the repl.

In the follwing I'd like to present the ideas I have for scripts and terms. Your ideas are welcome.
## 

I suggest adding a new command to the repl, which lets you set the reduction strategy.
For example:

```
λ> :strategy normal
The reduction strategy "normal" will now be used to evaluate terms.
λ>
```
## 

As for the evaluation my idea is to add a flag to the interpreter.

rexim:
The problem is that our Smart Church Structure Detector completely relies on the normal order.

For example, here I just hardcoded the REPL to use call-by-name instead of normal order:

``` morganey
λ> (SUCC 0)
term: (λ f . (λ x . (f (((λ f . (λ x . x)) f) x))))
λ>
```

And here you go! The result is not recognizable as a number anymore! From the call-by-name's point of view there are no beta-redices in the head position of the term so nothing to reduce. But the normal order actually goes deeper and reduces everything even under the lambda abstractions.

I'm ok with this feature only if
1. The normal order is the default strategy always;
2. We have properly documented the limitation of our Smart Detector;
3. We have some sort of a warning in the REPL about possibly broken Smart Detection when the user changes the default strategy;

keddelzz:
I didn't think that the detection of terms will be a problem, if they're evaluated using the call-by-name reduction strategy.

Actually, I like the detection of terms.
So maybe this isn't a good idea, should I close this issue then?

rexim:
No-no, let's keep it. It's actually interesting. I'm just thinking, is it possible to craft the `succ` operation that reduces to the normal form with call-by-name strategy? I don't actually know the answer. :) It's probably possible. And if it is, the implementation of such function is gonna be completely different. I mean in that case the detection problem is actually just a matter of using right implementations. We need to research that. So let's keep this issue opened. :)

keddelzz:
Ok! I'll add the research label :)

rexim:
Cool! Thanks!

PR #298, opened 2016-11-08
Added turing machine example

BjoernLoetters:
Finally I had the time to write a turing machine in Morganey!

Theoretically it can execute any turing machine program (and thus shows that Morganey is turing complete!). 

Unfortunately the morganey program itself has a runtime complexity which is nearly exponential. This behavior is the result of the evaluation strategy "call-by-name" Morganey uses. It could be fixed by using the "call-by-need" strategy. 

But where does this come from? Every time a turing machine is executed in a particular state (lets call it state "n"), it also executes the machine in the previous state ("n - 1"). Because parameters are evaluated each time they are used, the machine in state "n" executes the machine in state "n - 1" not only once, but a few times. The machine in state "n - 1" does the same and so on. As one can imagine, the algorithm builds up a tree of turing machines.

The evaluation strategy "call-by-need" would fix that problem, because once the machine in state "n - 1" is calculated it would never have to be calculated again. So the machine in state "n" would calculate the machine in state "n - 1", where the machine in state "n - 1" would calculate the machine in state "n - 2" and so on.  If I've not messed up, this should reduce the runtime complexity to a linear behavior.

In contrast, executing a single step is fast. So it is possible to execute a single step, overwrite the initial turing machine with the newly calculated one and run it again and again.

I've added a brief explanation of the usage to the file.

### DIFF
--- a/docs/samples/turing-machine.mgn
+++ b/docs/samples/turing-machine.mgn
@@ -110,18 +110,17 @@ def getTape := \machine.head machine
 /** returns the element at the n-th position */
 def getN := Y (\rec.index.list. (gt index 0) (rec (minus index 1) (tail list)) (head list))
 
-/**
- * Example usage:
- *
- * def tape := []
- * def program := [
- *   [1, 0, 1, 1, 2],
- *   [1, 2, 'x', 1, 0],
- *   [2, 0, 2, 1, 3],
- *   [3, 0, 1, 0, 1]
- * ]
- * def head := 0
- * def start := 1
- * def machine := newMachine tape head start program
- * def main := runMachine machine 
- */
+/** Example usage: */
+/*
+def tape := []
+def program := [
+  [1, 0, 1, 1, 2],
+  [1, 2, 'x', 1, 0],
+  [2, 0, 2, 1, 3],
+  [3, 0, 1, 0, 1]
+]
+def head := 0
+def start := 1
+def machine := newMachine tape head start program
+def main := runMachine machine
+*/
--- a/docs/samples/turing-machine.mgn
+++ b/docs/samples/turing-machine.mgn
@@ -1,0 +1,127 @@
+load std.prelude
+
+/**
+ * Format of the Turing Machine:
+ * [tape, head, state, program]
+ *
+ * - tape = list of numbers
+ * - head = the current read/write position (initially 0)
+ * - state = the current state of the machine (initially 1)
+ * - program = the program of the turing machine
+ *    - represented by a list of lists
+ *    - each inner list is a 5-tuple consisting of:
+ *      1. the current state
+ *      2. the character to read
+ *      3. the character to write
+ *      4. the move direction (0 = left, 1 = right)
+ *      5. the next state
+ *
+ * The state 0 is the halt state.
+ */
+
+/** creates a new turing machine based on the given tape, head, state and program */
+def newMachine :=
+  \tape.head.state.program.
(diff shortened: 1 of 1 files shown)


rexim:
@kuchenkruste this is so cool!! :)

I'll try to take a look later today. Thanks!

Newlifer:
Awesome!

rexim:
> This behavior is the result of the evaluation strategy "call-by-name" Morganey uses.

Your assumption is wrong. Morganey doesn't use call-by-name evaluation strategy. It uses normal order. Please see #84 for more info on that.

docs/samples/turing-machine.mgn
```diff
@@ -0,0 +1,127 @@
...
+ *   [1, 0, 1, 1, 2],
+ *   [1, 2, 'x', 1, 0],
+ *   [2, 0, 2, 1, 3],
+ *   [3, 0, 1, 0, 1]
+ * ]
+ * def head := 0
+ * def start := 1
+ * def machine := newMachine tape head start program
+ * def main := runMachine machine 
+ */
```
rexim: I'd suggest to organize the comments in this section like:

``` c
/** Example usage: */
/*
def tape := []
def program := [
  [1, 0, 1, 1, 2],
  [1, 2, 'x', 1, 0],
  [2, 0, 2, 1, 3],
  [3, 0, 1, 0, 1]
]
def head := 0
def start := 1
def machine := newMachine tape head start program
def main := runMachine machine 
*/
```

It makes it easier to uncomment even in the most minimalistic text editor in the world.

rexim (changes requested):
How long is the example at the bottom is supposed to work? I just uncommented it, ran with `sbt "run docs/samples/turing-machine.mgn"` and waited for ~10 minutes. The program was hanging.

Linux, Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz

Was it ok on your system?

docs/samples/turing-machine.mgn
BjoernLoetters: I'll do that by tomorrow.

BjoernLoetters:
I've changed the example usage comment. 

> How long is the example at the bottom is supposed to work?

To be honest: I don't know. So far only machines with `<= 2` states terminated on my PC. But you can execute the example step by step: 
1. Replace `def main := runMachine machine` with `def main := executeStep machine` 
2. Overwrite the machine `machine` with the computed machine (this is the machine which is in state `n + 1`)
3. Repeat step 1. and 2. as long as the state of the machine is not `0`

> > This behavior is the result of the evaluation strategy "call-by-name" Morganey uses.
> 
> Your assumption is wrong. Morganey doesn't use call-by-name evaluation strategy. It uses normal order. 

Probably I've messed something up - but the point is, that as long as morganey evaluates parameters (i.e. the expressions which are bound to them) each time they are used, the runtime won't get any better.

rexim:
> To be honest: I don't know. So far only machines with `<= 2` states terminated on my PC

Cool! Thanks for the information.

> Probably I've messed something up - but the point is, that as long as morganey evaluates parameters (i.e. the expressions which are bound to them) each time they are used, the runtime won't get any better.

Got it thanks!

rexim (approved):

rexim:
@kuchenkruste thank you very much for such an awesome sample! :)

A bit later I'll create an issue to address the performance concern. I think it should be discussed separately.

PR #302, opened 2016-11-20
Refactor ReplAutocompletion

### LINKED ISSUE #284: Refactor ReplAutocompletion
I suggest refactoring `ReplAutocompletion` before #257.

Currently `ReplAutocompletion` implements a few functions, which should be moved to `ModuleFinder` before #257. Those Functions all deal with finding and listing module-files and should therefore be placed in `ModuleFinder`.
##### [topLevelDefinitions](https://github.com/morganey-lang/Morganey/blob/82d3165bdc4943b835d2e0f61765327de61d8215/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala#L54)

``` scala
/** 
  * Returns a sequence of all directories and .*mgn-Files at the top level of all module paths 
  *
  * Imaging having a two folders in the morganey path `modsA` and `modsB`:
  *
  * `modsA`:
  * +---list.mgn
  * +---pair.mgn
  * |
  * \---math
  *     \---arithmetic.mgn
  *
  * `modsB`:
  * +---logic.mgn
  * +---bools.mgn
  * |
  * \---numbers
  *     +--- floats.mgn
  *     \--- whole-numbers.mgn
  *
  * In that case `topLevelDefinitions` should return
  * {{{
  * Seq(list.mgn, pair.mgn, math, logic.mgn, bools.mgn, numbers)
  * }}}
  */
def topLevelDefinitions(): Seq[File] = ???
```
##### [findAllModulesIn](https://github.com/morganey-lang/Morganey/blob/82d3165bdc4943b835d2e0f61765327de61d8215/src/main/scala/me/rexim/morganey/ReplAutocompletion.scala#L58)

``` scala
/**
  * Returns a list of tuples (a, b), where 
  * a = the root directory of this module-path
  * b = the path of a certain module
  * 
  * Requirement: `b` is a sub-path of `a`
  */
def findAllModulesIn(path: String): Seq[(File, File)] = ???
```

If we move those two functions from `ReplAutocompletion` to `ModuleFinder` and generalize them for `java.io.File` and `java.net.URI` somehow, #257 should be much easier to implement.

rexim:
### Checklist

- [X] ~Update docs~
- [X] Add/Update Unit Tests

---

Close #284

### DIFF
--- a/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
+++ b/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
@@ -17,6 +17,9 @@ object ModuleFinder {
   def relativeFileToLoadPath(relativeFile: String): String =
     relativeFile.replace(File.separatorChar, '.')
 
+  def validMorganeyElement(f: File) =
+    f.isDirectory || isMorganeyModule(f)
+
   def isMorganeyModule(file: File) =
     file.isFile() && (file.getName endsWith s".$fileExtension")
 }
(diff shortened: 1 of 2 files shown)


src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
```diff
@@ -34,4 +37,47 @@ class ModuleFinder(val paths: List[File]) {
     val resourceUrl = classLoader.getResource(resourcePath)
     Option(resourceUrl)
   }
+
+  /**
+    * Returns a sequence of all directories and .*mgn-Files at the top level of all module paths
+    *
+    * Imaging having a two folders in the morganey path `modsA` and `modsB`:
```
ForNeVeR: I'm not a language expert, but shouldn't `Imaging` be `Imagine` here?

ForNeVeR (changes requested):

src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
keddelzz: I think so, too :)

rexim: Fxd

keddelzz (approved):

ForNeVeR (approved):

rexim:
@keddelzz @ForNeVeR thanks for the review!

PR #308, opened 2016-11-27
Find all modules in Morganey indices

### LINKED ISSUE #305: Make ModuleFinder collect and merge indices
Split from #257

rexim:
Close #305

### DIFF
--- a/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
+++ b/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
@@ -40,7 +40,7 @@ class ModuleFinder(val paths: List[File]) {
     Option(resourceUrl)
   }
 
-  def findAllModulesInClasspath(): Set[String] = {
+  def findAllModulesInIndex(): Set[String] = {
     import scala.collection.JavaConversions._
 
     val classLoader = this.getClass.getClassLoader
--- a/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
+++ b/src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
@@ -5,6 +5,8 @@ import java.net.URL
 
 import ModuleFinder._
 
+import scala.io.Source
+
 object ModuleFinder {
   val fileExtension = "mgn"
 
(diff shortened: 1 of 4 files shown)


src/main/scala/me/rexim/morganey/module/ModuleFinder.scala
```diff
@@ -38,6 +40,16 @@ class ModuleFinder(val paths: List[File]) {
     Option(resourceUrl)
   }
 
+  def findAllModulesInClasspath(): Set[String] = {
```
rexim: I think it's better to rename it to something like `findAllModulesInIndex` to emphasize that it's not the only way of discovering modules. It is also possible to access a module if you know its full path, even if such module is not registered in the Morganey index.

ForNeVeR (changes requested):
I think that's a good moment to add some tests for this function. If you prefer to not mess with classpath, you could use mocked ClassLoader or something like that. Probably API of this function needs to be changed (I think that's a good idea to pass `ClassLoader` as a parameter to test it easier).

rexim:
@ForNeVeR wow! I didn't even think about mocking ClassLoader for testing! I feel so dumb now! :D Thank you very much!

If we are able to mock ClassLoader for testing we don't need all of that bullshit with filtering module containers that I suggested in the description of this PR. I'm gonna address that a bit later. Thanks!

keddelzz (approved):

rexim:
@ForNeVeR @keddelzz @kuchenkruste quick question. Which mocking framework would you use for Morganey and why?

ForNeVeR:
@rexim I actually have no experience with mocks in Scala, so I'd go with ScalaMock, because it have an obvious name :)

rexim:
My personal answer on the above question is mockito, because I already know it, it just works and ScalaTest supports it in a form of [MockitoSugar](http://doc.scalatest.org/1.8/org/scalatest/mock/MockitoSugar.html).

rexim:
If no additional answers are provided within two hours starting from now I'm applying mockito and we are going with it. Just wanna close this one before the stream. :)

rexim:
Time's up! Thanks everyone! The PR is ready for the review.

ForNeVeR (approved):

rexim:
@ForNeVeR @keddelzz thank you guys for the review!

I guess appveyor/pr is gonna hang again, so I'm mergin this one.

PR #335, opened 2017-03-04
Implement Module.{bindings, dependencies} 

### LINKED ISSUE #329: Implement Module.{bindings, dependencies}
Split from #320

https://github.com/morganey-lang/Morganey/blob/458946d8fefc71fffc8c98834a7752c9081925ab/src/main/scala/me/rexim/morganey/module/Module.scala#L10

rexim:
Close #329

### DIFF
--- a/src/main/scala/me/rexim/morganey/module/Module.scala
+++ b/src/main/scala/me/rexim/morganey/module/Module.scala
@@ -1,13 +1,35 @@
 package me.rexim.morganey.module
 
-import me.rexim.morganey.ast.MorganeyBinding
+import me.rexim.morganey.ast._
+import me.rexim.morganey.syntax._
+import me.rexim.morganey.util._
 import scala.util._
+
+import java.net.URL
 
 class Module(modulePath: ModulePath, classLoader: ClassLoader = Module.getClass.getClassLoader) {
   def canonicalPath: String =
     modulePath.asCanonicalPath.path
 
-  // TODO: Implement Module.{bindings, dependencies}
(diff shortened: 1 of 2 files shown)


src/main/scala/me/rexim/morganey/module/Module.scala
```diff
@@ -1,13 +1,35 @@
...
+import java.net.URL
+
 class Module(modulePath: ModulePath, classLoader: ClassLoader = Module.getClass.getClassLoader) {
   def canonicalPath: String =
     modulePath.asCanonicalPath.path
 
-  // TODO: Implement Module.{bindings, dependencies}
-  def bindings: Try[Set[MorganeyBinding]] = ???
-  def dependencies: Try[Set[Module]] = ???
+  private def nodes: Try[List[MorganeyNode]] = {
```
ForNeVeR, line 14: Is it possible to make it `lazy val nodes`?

ForNeVeR (approved):

src/main/scala/me/rexim/morganey/module/Module.scala
rexim: Yeah, I think it is, but my trail of thoughts was the following: `lazy val` essentially caches the module content. If we change it the module object will still return the "cached" values and to reload the module we have to create the same instance again.

ForNeVeR: Alright, let's leave it as-is.

keddelzz (approved):

rexim:
@ForNeVeR @keddelzz thank you so much guys for the review!

